PR #6, opened 2011-07-22
a test for request events

ghost:
This new test fails since requests are getting emitted twice for events. This is fixed in my tap-harness patch.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,2 +1,8 @@
-Isaac Z. Schlueter <i@izs.me> (http://blog.izs.me)
+# contributors sorted by whether or not they're me
+Isaac Z. Schlueter <i@izs.me>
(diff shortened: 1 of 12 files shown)


isaacs:
So, the whole thing with "Results" emitting "result" events is going to take a bit of a more significant refactor, I'm afraid.

With your patch, it doesn't work with nested tests, like this:

``` javascript
test("parent test", function (t0) {
  t0.plan(2)
  t0.once("result", function (res) {
    t0.ok(res.ok, "It's ok here, too")
  })
  t0.test("child test", function (t1) {
    t1.ok(true, "It's ok")
    t1.end()
  })
})
```

I'm going to need to take a closer look at how tests, harnesses, and result lists are wired together, maybe rewrite a few bits.

PR #12, opened 2011-10-12
Show syntax errors (stderr).

### LINKED ISSUE #4: Syntax errors result in tap being "mysterious"
A syntax error at the top of my file resulted in output like this (basically it tells you nothing):

bluesnoop:node-asn1 mark$ ./node_modules/.bin/tap ./tst
ok ber_decoder.test.js .............. 0/0
total ............................... 0/1

not ok
bluesnoop:node-asn1 mark$ TAP=1 ./node_modules/.bin/tap ./tst
# ber_decoder.test.js

not ok 1 ber_decoder.test.js

1..1
# tests 1
# fail  1

bluesnoop:node-asn1 mark$

smh:
Not sure if this is the best way to do it, but this will show stderr when node finds syntax-errors in the test-script.
Should fix #4

### DIFF
--- a/bin/tap.js
+++ b/bin/tap.js
@@ -24,6 +24,10 @@ if (process.env.TAP || process.env.TAP_DIAG) {
       console.log("    " + TapProducer.encode(details.list)
                   .split(/\n/).join("\n    "))
     }
+    if (results.stderr && details.ok && details.tests.pass === 0) {
+      // perhaps a compilation error or something else failed...
+      console.error(results.stderr)
+    }
   })
   r.on("end", function () {
     //console.log(r)


isaacs:
That's pretty clever, and should avoid any kind of interference with the tap output, since tap-formatted output is always on stdout.

Seems like it should maybe live in tap-runner instead, though.  What do you think?

smh:
I agree, it fits better in the tap-runner.

I just tried that, and one issue I found was that the stderr output in this case is now shown before the tap-result, rather than after. Since this should not interfere with the tap-output, and in fact is just information to the user, I don't suppose that matters.

I will send a pull request on the tap-runner instead.

PR #15, opened 2011-11-12
add -d, --debug option to pipe the test's stderr

dominictarr:
resolves this issue: https://github.com/isaacs/node-tap/issues/3

I've only supported piping stderr, because that is where debugging output should go. 
Maybe people will still want to show stdout, but then you would have to filter out what tap-consumer did not parse. 

And that isn't coming in on this sunday morning pull request.

### DIFF
--- a/bin/tap.js
+++ b/bin/tap.js
@@ -1,14 +1,19 @@
 #!/usr/bin/env node
 
-var argv = process.argv.slice(2)
-  , path = require("path")
+var path = require("path")
   , Runner = require("tap-runner")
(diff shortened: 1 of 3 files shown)


jeffbski:
Isaac, there is also a similar pull request in issue 31 - https://github.com/isaacs/node-tap/pull/31

In my version, I did not make stderr optional, it always forwards stderr (if TAP and TAP_DIAG are both not set). 

I did add a test to verify the output if you want to adapt it for this patch. 

Whatever you prefer.

PR #22, opened 2011-12-05
similar() and dissimilar() bug fix when `null` is used

jhs:
Today `t.throws()` was giving me trouble. This is not the first time. Here is the problem:

```
var expected = {"name":"MyError", "message":"The message"};
t.throws(function() {
  console.log("I do not throw after all!")
}, expected);
```

Ultimately this is converted to `assert.similar(null, expected)` which calls `selectFields(null, expected)` which... wait for it... throws an exception trying to loop through `null`'.

These patches fix that. The only weird thing is that now `assert.similar()` runs two assertions internally. Not sure if that throws anything off such as the auto-incrementing `id` or whatever. All the tests still pass at least.

P.S. Thanks very much for merging all these into one repository!

### DIFF
--- a/lib/tap-assert.js
+++ b/lib/tap-assert.js
@@ -239,7 +239,8 @@ function dissimilar (a, b, message, extra) {
   if (extra && extra.skip) return assert.skip(message, extra)
   // test that a has all the fields in b
   message = message || "should be dissimilar"
-  return inequivalent(selectFields(a, b), b, message, extra)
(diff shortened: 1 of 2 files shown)


autarch:
An alternate fix is to add something like this in the throws method:

```
if (wanted && !found) {
  found = {}
}
```

isaacs:
Landed in master.  Thanks!

PR #28, opened 2011-12-07
Fix #24: ensure separate test cases get their own 'conf' copy

### LINKED ISSUE #24: tap harness caches the test name on a possibly shared 'conf' object
Code ("foo.js"):

```
var config = {timeout: 8000}
test("one", config, function(t) {
  t.ok("one");
  t.end()
});
test("two", config, function(t) {
  t.ok("two");
  t.end();
});
test("three", config, function(t) {
  t.ok("three");
  t.end();
});
```

Output:

```
$ TAP=1 ./node_modules/.bin/tap foo.js
# three
ok 1 (unnamed assert)
# three
ok 2 (unnamed assert)
# three
ok 3 (unnamed assert)
# tests 3
# pass  3
# ok
ok 4 foo.js


1..4
# tests 4
# pass  4

# ok
```

Notice "# three" for each test.

One fix is to create a copy of the 'conf' object: https://gist.github.com/1435063

Alternatively, if only the 'name' is put on `this.conf`, then could just have a separate `this.name`.

trentm:


isaacs:
Merged into master, published on 0.1.1

PR #31, opened 2011-12-09
runner forwards stderr logging to stderr output

jeffbski:
If a user logs to stderr (like console.error) from code or a test,
tap runner should forward this on to the tap runner stderr output.

Test included.

### DIFF
--- a/bin/tap.js
+++ b/bin/tap.js
@@ -24,6 +24,7 @@ if (process.env.TAP || process.env.TAP_DIAG) {
       console.log("    " + TapProducer.encode(details.list)
                   .split(/\n/).join("\n    "))
     }
+    if (results.stderr) console.error(results.stderr); //show stderr if exists 
   })
   r.on("end", function () {
     //console.log(r)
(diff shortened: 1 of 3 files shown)


jeffbski:
Pull request should have said, "Runner should forward stderr logging to stderr output".

Also this would only be true env variables TAP and TAP_DIAG are not set.

autarch:
Please, please apply this.

bigeasy:
This patch will allow stderr to go to the process stderr, correct?

jeffbski:
Yes, but only if environment variables TAP and TAP_DIAG are not set, which the tap runner treats differently (which I believe is when it needs to pipe it to something else for consumption. It isn't real clear in the code and no doc about it). If they are set, then I am assuming the tap output might be being consumed in a special way so I do not introduce stderr.

But normally, unless you set those environment values for specific purpose, stderr would be forwarded to the processes stderr. 

That's based on my assumptions from reading Isaac's code.

bigeasy:
Please apply this patch. During development, I like to add temporary `console.error` calls for debugging. Actually, I'm accustomed to `console.log` for debugging, but I understand that tap needs that to be TAP output. I like how this gets back to the conventions of using stdout for output and stderr for meta output.

PR #35, opened 2012-01-06
fixes deep equals

ghost:
`t.deepEqual()` and company return false negatives when the key order doesn't match for two non-Array objects.

### DIFF
--- a/lib/tap-assert.js
+++ b/lib/tap-assert.js
@@ -314,7 +314,18 @@ function selectFields (a, b) {
   return ret
 }
 
+function sortObject (obj) {
+  return Object.keys(obj).sort().reduce(function (acc, key) {
+    acc[key] = obj[key]
+    return acc
+  }, {})
+}
+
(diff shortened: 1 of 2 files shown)


jhs:
Love it! I'm a big deepEqual user myself. IMHO I prefer not to put failing tests in their own commits because `git bisect` might checkout that commit and your test suite will fail, complicating the job of fixing the _real_ bug.

ghost:
I like it better when failing tests have their own commits so I can more easily ditch an implementation change designed to pass the test without discarding the test.

isaacs:
Commits can be squashed or split easily enough.  Imo, it's best for pull reqs to reflect the details as clearly as possible (usually in many commits), and then (usually) get squashed into a single commit in the upstream (or merged with `--no-ff` if it's worth keeping the detail.)

isaacs:
merged and pushed to all the places.

PR #36, opened 2012-01-07
Fix #33

### LINKED ISSUE #33: ignore non executable scripts
tap attempts to run any thing in the test dir that is not .js as a script.

should ignore files that are not executable.

emberian:
tap-runner.js no longer tries to execute scripts if they don't have the executable bit set for anyone. Is kinda dirty as it doesn't check if the user can execute the file. I'll add that in if you want, I just don't have time right now.

Adds two tests, executed.sh and not-executed.sh. not-executed.sh will always fail, if it is run.

### DIFF
--- a/lib/tap-runner.js
+++ b/lib/tap-runner.js
@@ -143,13 +143,17 @@ Runner.prototype.runFiles = function (files, dir, cb) {
         }
 
         var cmd = f, args = [], env = {}
-
         if (path.extname(f) === ".js") {
           cmd = "node"
           args = [fileName]
         } else if (path.extname(f) === ".coffee") {
           cmd = "coffee"
           args = [fileName]
+        } else if ((st.mode & 0111) == 0) {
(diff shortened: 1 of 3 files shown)


isaacs:
Yes, can you please have it test against 0100, then 0010 if the gid matches and you can call process.getgid, then 1 if the uid matches and you can call process.getuid.  The getuid/getgid functions will be missing if the system doesn't support it.

Also, 2 space indent, not 4.

Other than that, +1.

emberian:
I wish I could add more tests but I really can't, not with the way git preserves permissions. Perhaps a pretest hook of some sort that creates files with various permissions (or does git save /which/ executable bit is set?)

emberian:
Bump

isaacs:
Landed in master, thanks!

isaacs:
Landed in master, thanks!

PR #42, opened 2012-02-22
deepEqual is still actually broken as evidenced by the extra test case I added to test/deep.js

ghost:
I swaped out the stringify() checking to use the deep-equal module which is just assert.deepEqual() that returns true or false instead of throwing. The comparing the strings doesn't seem necessary anymore since the diffs are computed by difflet.


isaacs:
landed like an eagle.

PR #46, opened 2012-05-11
buffer comparison

ghost:
This is a special case for comparing buffers since deepEqual will compare against the parent buffer which is probably unimportant for the overwhelming majority of buffer comparisons.

I also added `t.notSame()` since `t.same()` didn't have a buddy.


isaacs:
landed! published! thank you!

PR #48, opened 2012-07-18
Added support for -gc and --expose-gc

siddMahen:
- Exposes the `gc()` function in tap tests
- Added a test "expose-gc-test.js" to back this up

All tests pass.


siddMahen:
@isaacs feedback?

isaacs:
Landed on master/0.3.0

PR #49, opened 2012-07-19
IE9 compat

Raynos:


isaacs:
Landed the IE9 patch, not the trick npm patch, though.

Raynos:
@isaacs I forgot about the trick npm patch.

npm does aggressive caching when I set up `tap: git://github.com/Raynos/node-tap#branch` and making the version number  different tells it to stop doing that.

PR #52, opened 2012-07-27
End exception infinite loop

ghost:
This pull request fixes issue #50 where exceptions in `t.on('end', fn)` listeners cause an infinite loop.
I added a failing test and then fixed the issue with a one-line change in `Test.prototype.threw` to check if the test had already been ended before calling `t.end()`.


isaacs:
landed on master, thanks!

PR #58, opened 2012-09-19
require buffer explicitly so tap can be made to work in browsers

ghost:
I'm attempting to use tap in the browser but one of the tiny things left is how `Buffer` is a global in node but must be explicitly required in browserify for download-size reasons. This one-line change imports `Buffer` explicitly.


Raynos:
+1

isaacs:
landed in 0.3.3 thanks!

PR #59, opened 2012-11-06
Adding `assert.notStrictEqual()` for 'assert' compatibility

ELLIOTTCABLE:
Although node-tap's `assert` is already somewhat incompatible with Node's 'assert' module (for instance, `assert.equal()` here uses threequality instead of coercive equality), the complete lack of `notStrictEqual` to match 'assert's might as well be fixed.


isaacs:
Landed in v0.3.2 8884ec1572b55eba0b3dffe06b5e45e6dd9cda66 with slight edit.

PR #66, opened 2013-01-06
Add -g flag to installation instructions in README.md

MRdNk:
Installation instructions missing -g global installation flag.
'npm install tap' will not allow use of 'tap ./test'
- Unless I'm missing something.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Default Usage:
 2. Put a bunch of test scripts in there.  If they're node programs, then
    they should be ".js".  Anything else is assumed to be some kind of shell
    script, which should have a shebang line.
-3. `npm install tap`
+3. `npm install tap -g`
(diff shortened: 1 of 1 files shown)


isaacs:
Yes, you are missing that npm puts `node_modules/.bin` at the start of the PATH when running scripts.

The README should also mention that you should put `tap ./test` in your `scripts.test` in your package.json file, though.

MRdNk:
So this is scripts in package.json

```
"scripts": {
    "test": "tap test/*.js"
  }
```

generated by npm init

However...

```
tap ./test
```

Errors ... -bash: /usr/local/bin/tap: No such file or directory

It's fine if I run 'npm test'

isaacs:
Yep.  If you update the pull req to say to use package.json and `npm test`, then I'll take it.

PR #67, opened 2013-01-06
Updated to readme.md installation instructions

MRdNk:
Installation instruction amendments.
- Added an example gist (might be OTT).


isaacs:
Landed on 3ff1eef1b672bcff0c9dbd97f593fd9963ee072b.  Thanks!

PR #69, opened 2013-03-03
Call EventEmitter constructor in child classes.

wanderview:
Currently the `Harness` and `Results` classes inherit from `EventEmitter`,
but they do not call the `EventEmitter` constructor.

See joyent/node#4891.


isaacs:
Fixed another way, thanks.

PR #71, opened 2013-03-14
Pass debug and debug-brk flags to node for debugging Tap tests

dberesford:
Adds two new flags to tap to assist with debugging your node Tap tests:

--debug: Pass the '--debug' flag to node for debugging
--debug-brk: Pass the '--debug-brk' flag to node for debugging

e.g.
$ tap --debug-brk test/foo.js
debugger listening on port 5858

You can then fire up node-inspector and debug like a boss.

### DIFF
--- a/bin/tap.js
+++ b/bin/tap.js
@@ -16,6 +16,8 @@ var argv = process.argv.slice(2)
     , tap: Boolean
     , timeout: Number
     , gc: Boolean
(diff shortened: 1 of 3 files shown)


isaacs:
df142bbd0fe9be86b9c54340c2f15e1140200cea Thanks!

PR #72, opened 2013-03-25
fixes issue #70 - wrong args passed to path.resolve

cioddi:
In node 0.10 path.resolve only accepts string parameters but {array}.map passes the array index and the array itself as additional parameters.
I added a function to make sure path.resolve only gets string parameters.

fixes: https://github.com/isaacs/node-tap/issues/70


isaacs:
Landed on master, and published to npm as 0.4.1.  Thanks!

PR #75, opened 2013-04-23
tests: segv actually segfaults

tchollingsworth:
The 'segv' test is written to expect SIGBUS, even though it seems to actually segfault as its name would suggest.

This patch makes it expect SIGSEGV.

### DIFF
--- a/test/segv.js
+++ b/test/segv.js
@@ -38,7 +38,7 @@ test('segv', function (t) {
           'name': ' ././segv',
           'exit': null,
           'timedOut': true,
-          'signal': 'SIGBUS',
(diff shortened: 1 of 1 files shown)


joliss:
I'm seeing the same problem on Ubuntu 13.04. This patch fixes it for me.

joliss:
Actually, the test turns out to be flickering: Mostly I get SIGSEGV, but every once in a while I'm getting SIGTERM: https://gist.github.com/joliss/495ed06c9d0192d4c62f

Perhaps the test should be fuzzy and not care about the kind of signal?

isaacs:
Fixed already

PR #76, opened 2013-04-27
Use setImmediate when available.

aredridel:


isaacs:
I don't get it.  Why is setImmediate better in this case?  (Or, perhaps, why _was_ it better 7 months ago, and is that still the case?)

aredridel:
I was getting recursive nextTicks without this.

isaacs:
doesn't merge cleanly.  Should be fine on modern node anyway.

PR #77, opened 2013-05-13
Tap hangs with Node 0.6

joliss:
With Node v0.6.21, tap hangs for 30 seconds exactly and then exits with exit code `0` without any output:

```
node-tap $ bin/tap.js test/*.js
node-tap $ echo $?
0
```

Node 0.6 ships with Ubuntu 13.04 (if you don't add Chris Lea's PPA). Is Node 0.6 still worth supporting, or has everyone moved on by now?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,9 @@
   "description": "A Test-Anything-Protocol library",
   "bin": "bin/tap.js",
   "main": "lib/main.js",
+  "engines": {
+    "node": ">=0.8"
+  },
   "dependencies": {
     "inherits": "*",
     "yamlish": "*",


joliss:
(I got bitten by this when I added Node 0.6 on Travis for my own library, and it was false-green: https://travis-ci.org/joliss/js-string-escape/jobs/7107588 )

isaacs:
> Is Node 0.6 still worth supporting, or has everyone moved on by now?

Node 0.6 is pretty dead at this point.  It's best to upgrade at least to 0.8, and there are basically zero downsides to doing so.

I'm totally ok with accepting a reasonable pull req to fix this, but I don't have a lot of time to do it myself.  My best guess is that it's waiting on a "close" event that should be an "end" event or something.

joliss:
Sounds good. Should we specify Node >=0.8 in the package.json then? I've attached a commit to this issue.

isaacs:
Landed on 7b39beb1e23cb7b033f90e2a103924a21cfbb901.  Thanks!

PR #78, opened 2013-05-18
Add cmd line args to enable harmony features and enforce strict mode

ttaubert:
Just like support for --expose-gc has been added quite a while ago it would be great to support --use-strict and --harmony as well. ES6 finalization is getting closer and V8 recently implemented generator support that will be available in node 0.11.2 (well, with --harmony).

While working on a small library making use of generators I figured that it would be great to simply be able to use 'npm test' locally and also have tests run in strict mode with harmony features on Travis. Travis said they're upgrading their machines to node 0.11.2 soon.

### DIFF
--- a/bin/tap.js
+++ b/bin/tap.js
@@ -16,6 +16,8 @@ var argv = process.argv.slice(2)
     , tap: Boolean
     , timeout: Number
     , gc: Boolean
+    , strict: Boolean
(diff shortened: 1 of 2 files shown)


isaacs:
af808bd2fcbc312149d28e67b6333150e377a29e.  Thanks

PR #82, opened 2013-06-02
Add `fulfill` event to listen for when the `plan` count is met.

chbrown:
I want to test several actions (say, 5 of them) against a single http server, but I don't want to nest them and I don't really need complex async semaphore. So I set `t.plan(5)` and then send off each of my requests, each of which have an async callback doing something like `t.equal("Something", res.body)`. I want to catch when those are all done, so that I can run `server.close()` and return. This event, `fulfill`, is triggered kind of in between the last `result` event and the `end` event (which doesn't occur until the child process ends).

### DIFF
--- a/lib/tap-test.js
+++ b/lib/tap-test.js
@@ -62,6 +62,8 @@ Test.prototype.result = function (res) {
   this._testCount ++
   this.emit("result", res)
   if (this._plan === this._testCount) {
+    // plan has been met
+    this.emit("fulfill")
     process.nextTick(this._endNice.bind(this))
   }
 }


isaacs:
Why don't you just do:

``` javascript
tap.tearDown(function() {
  server.close();
});
```

in your test file?

Alternatively, you could do this for more control and logging:

``` javascript

var server;
test('setup', function(t) {
  server = http.createServer(function(req, res) { blerg });
  server.listen(PORT, function() {
    t.pass('server listening');
    t.end();
  });
});

test('do requests and stuff', function(t) {
  t.plan(5);
  // make 5 requests or whatever
});

test('teardown', function(t) {
  server.close(function() {
    t.pass('server closed');
    t.end();
  });
});
```

I think adding a 'fulfill' event is unnecessary.

chbrown:
Oh, totally did not know about `t.tearDown()` -- that's perfect. Thanks.

PR #84, opened 2013-06-05
NaN is equal to NaN

### LINKED ISSUE #83: NaN as a special case
It would be nice if NaN was treaded as a special case, so the following would be OK:

``` javascript
t.deepEqual({ prop: NaN }, { prop: NaN });
```

Right now the diff says:

```
diff:    |
  {
    "prop" : NaN, // != NaN
  }
```

and that is just a bit odd.

AndreasMadsen:
Fixes #83

This changes:
- t.deepEqual
- t.equal
- t.like

and there opposites

### DIFF
--- a/lib/tap-assert.js
+++ b/lib/tap-assert.js
@@ -1,6 +1,6 @@
 // an assert module that returns tappable data for each assertion.
 var difflet = require('difflet')
-  , deepEqual = require('deep-equal')
+  , deepIs = require('deep-is')
   , bufferEqual = require('buffer-equal')
   , Buffer = require('buffer').Buffer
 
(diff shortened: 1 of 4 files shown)


othiym23:
This is an oddity of IEEE754, but that doesn't mean that `tap` should paper over it unless a developer specifically wants that behavior. IMHO this shouldn't be the default.

isaacs:
I'd be ok with adding `t.is()` and `t.deepIs()` methods, but changing `equal` is not a good idea.

isaacs:
Actually, `like()` is probably fine to change in this way.

pc-dedelhart:
Can't you do an isNaN() around the result and compare the two?

PR #92, opened 2014-01-07
fix; command field in tap output not valid

wavded:
Currently this fails:

```
not ok 1 test/settings/settings-test.js
  ---
    exit:     143
    timedOut: true
    command:  "node" "settings-test.js"
  ...
```

Because it cannot parse the "command" value, relevant yaml parsing error is:

> expected &lt;block end&gt;, but found Scalar

This fix changes the command value to be one quoted string instead which is
valid:

```
not ok 1 test/settings/settings-test.js
  ---
    exit:     143
    timedOut: true
    command:  "node settings-test.js"
  ...
```

Tests were ran against [InstantTap](http://instanttap.appspot.com/)

### DIFF
--- a/lib/tap-runner.js
+++ b/lib/tap-runner.js
@@ -296,7 +296,7 @@ function runFiles(self, files, dir, cb) {
 
           // tc.results.ok = tc.results.ok && ok
           tc.results.add(res)
-          res.command = [cmd].concat(args).map(JSON.stringify).join(" ")
+          res.command = '"'+[cmd].concat(args).join(" ")+'"'
           self.emit("result", res)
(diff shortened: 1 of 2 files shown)


wavded:
For some more context @isaacs and co, this issue came about using the [TAP Plugin](https://wiki.jenkins-ci.org/display/JENKINS/TAP+Plugin) for Jenkins.  Whenever we'd get a failing test that included the "command" it would throw away the whole report (due to the parsing error mentioned above).  We have been running this patch and it has been working for us.

isaacs:
3c2794e4b3d7523ce797afed520e04f0fde66060 Thanks!

PR #93, opened 2014-01-07
lib: don't assume 'node' is on the path

bnoordhuis:
Use process.execPath to start the test runner.  Assuming there is a
node binary on the path somewhere will break when there isn't one or
when it's the wrong version.


isaacs:
Thanks!  Landed and published on 0.4.7

PR #94, opened 2014-01-10
TAP: ignore skip & todo entries in yamlish data

macbre:
Don't include skip and todo entries in yamlish details of TAP


isaacs:
Can you add a test for this?

rmg:
@macbre ping

PR #95, opened 2014-01-21
Adds support for --exec (-e), specifying executable runner

tcr:
Adds an -e flag, forcibly setting the executed command (regardless of the extension of the file, .js, .coffee, or nothing). This allows for example running an alternative JavaScript interpreter, as well as other commands.


isaacs:
Not opposed to the feature.  Needs a test, though.

PR #98, opened 2014-04-17
Make SKIP and TODO more verbose

rmg:
Makes TAP reports more useful by allowing TODO and SKIP items to be annotated.

Also includes a small tweak to make the tests pass.


rmg:
Rebased since one of the commits was replaced with 44f5bcf148dbc47e9f74f742d96de8bcd06a75d4

rmg:
Now with 1000% more tests!

isaacs:
Landed on master.  Thanks!

PR #99, opened 2014-04-17
runner: Support test/test-* including non-js files

rmg:
When falling back on the "is it executable" test for files that don't
end in .js or .coffee, files that are determined to be executable as
tests should be referenced by a more complete path because $CWD is
highly unlikely to be included in $PATH.

The end result is `tap tests/test-*` works when the tests are a mix of JS and shell scripts.

### DIFF
--- a/lib/tap-runner.js
+++ b/lib/tap-runner.js
@@ -209,6 +209,7 @@ function runFiles(self, files, dir, cb) {
           } else if ((st.mode & 0001) == 0) {
             return cb()
           }
+          cmd = path.resolve(cmd)
         }
 
         if (st.isDirectory()) {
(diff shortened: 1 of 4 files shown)


rmg:
@isaacs you partially covered this with 44f5bcf. Something wrong with the test fix and the non-js pieces of this PR?

rmg:
Rebased/squashed to remove conflict.

rmg:
@isaacs bump.

I've got half a dozen repos using "strongloop-forks/node-tap" in their devDependencies because of this +3/-1 change. Is there anything I can do to get this merged?

rmg:
Now with 200% more tests! (actually, just enabled the ones that were already there but were never exercised)

isaacs:
Landed on master!

rmg:
Awesome, thanks! This will let me remove a git dep from a bunch of modules :-)

PR #101, opened 2014-07-08
Fixes to allow tap to run under --use_strict

macmenot:


isaacs:
Thanks for the nudge.  Landed but with a test and using `parseInt` instead of decimal numbers with a comment.

PR #103, opened 2014-08-13
Allowed nested async tests when plan announced

### LINKED ISSUE #102: async nested test harnesses (zalgo issue?)
Currently, the `process` function will automatically call `this.end()` whenever there are no children left to process in a harness. This is fine for synchronous tests where all the child harnesses are defined before `this.process` is first called on the `process.nextTick`. 

However, if you have a circumstance where you want to create child harness asynchronously, then `this.process` will see that there are no children left and call `this.end` automatically before your next child harness can be registered. 

I'm happy to submit a pull request to fix this, but I wanted to see how others think this should be resolved. IMHO, the harness should only end if `this.end` is called explicitly by some outside function or when the plan count is reached. Merely adding a conditional here https://github.com/isaacs/node-tap/blob/a444c755a981997d8934648b3e165a738b07780d/lib/tap-harness.js#L90 in `Harness.prototype.process` that only calls `this.end` if `this._plan && this._testCount === this._plan` should be sufficient AFAICT, but I will confirm this.

andrewdeandrade:
This patch allows you to asynchronously define nested test harnesses if you first announce a plan count. e.g. If I say that 3 assertions are expected and one of the assertions is nested test set, that nested test set can be declared asynchronously, so long as it is defined and finishes before the timeout. 

Fixes #102

### DIFF
--- a/lib/tap-harness.js
+++ b/lib/tap-harness.js
@@ -24,7 +24,7 @@ function Harness (Test) {
   this._Test = Test
   this._plan = null
   this._children = []
-  this._started = false
+  this._processing = false
 
   this._testCount = 0
   this._planSum = 0
@@ -40,7 +40,7 @@ function Harness (Test) {
 
   var p = this.process.bind(this)
   this.process = function () {
-    this._started = true
+    this._processing = true
     process.nextTick(p)
   }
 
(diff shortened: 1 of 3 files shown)


isaacs:
Fixed on master by cb3e353504aaaa98e1cecd1a86099bf2bdaae5c9.  Thanks!

PR #111, opened 2015-02-09
Fixed tests in 0.12 / iojs

### LINKED ISSUE #108: Tests fail in node 0.11
I'm currently looking into bringing 0.11 support to [testem](https://github.com/airportyh/testem) and the last remaining issues are our tap tests.

Currently node-tap doesn't pass (239/241) when running `npm test` on v0.11.15 . Is node 0.11 planned to be supported?

PS: Having a publicly available CI (like travis) to see, which nodes are currently tested and passing, would be awesome.

johanneswuerbach:
0.12 and iojs are printing “Debugger” instead of “debugger”

Fixes https://github.com/isaacs/node-tap/issues/108


isaacs:
Thanks!

PR #112, opened 2015-02-11
Don't fail suite if a test skips

### LINKED ISSUE #110: tap exits with failure on skipped tests
https://github.com/isaacs/node-tap/commit/d3ed42634288278d048972a8fe269f9c404e59c1 was quite helpful, except it doesn't deal with skipped tests.

```
1..8
# tests 8
# pass  6
# skip  2
% echo $?
1
```

So now tap runner says a test has failed if tests are skipped, which I think is the opposite of the intention of skipping a test. I don't have to skip it if its going to cause the tests to fail, skipped or not.

rmg:
Skipped tests aren't really pass or fail, so they shouldn't cause the suite to fail.

Resolves #110

### DIFF
--- a/lib/tap-results.js
+++ b/lib/tap-results.js
@@ -62,7 +62,7 @@ Results.prototype.add = function (r, addToList) {
       // console.error("Bailing out in result")
       this.bailedOut = true
     }
-    this.ok = !!(this.ok && r.ok)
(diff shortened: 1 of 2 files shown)


isaacs:
Landed on master, thanks!

PR #113, opened 2015-02-11
Update debugger test to ignore debugger message case

rmg:
Alternate to #111 for fixing #108


isaacs:
Landed #111, which is fine. Thanks for the choice, though :)

rmg:
I fixed it myself locally before noticing the other one, so it's all good :-)

PR #114, opened 2015-02-12
Update timeout signal test for node > 0.10

rmg:
Looks like newer versions of Node cleaned up this platform difference.


isaacs:
Landed on master, thanks!

PR #122, opened 2015-02-26
Improve consistency of TODO and SKIP directives

rmg:
- don't print 'undefined' for unexplained skip/todo
  
  Support for directive explanations was added in #98 but it was only
  being applied to TAP input, which created a blank explanation if one
  was not provided. The results from tap-assert did not ensure the
  explanation was a string, so "undefined" started popping up for TODO
  and SKIP tests.
  
  Also make sure we preserve assert.skip explanations if they are given,
  since skipped tests don't get the same "clone extra" treatment that
  other asserts do.
- don't count TODO failures as fatal
  
  This builds on the behaviour and test introduced in #112, extending
  it to cover TAP's other directive: TODO.
  
  The TAP 13 spec says a harness should not consider a set of tests as
  failing overall if the only failures are marked as TODO.
  
  What is unclear, however, so the exit code is left as is, representing
  the overall harness result.


rmg:
@isaacs more non-doc changes, so I won't merge until you've given the thumbs up.

isaacs:
:+1:

rmg:
Landed on master :-)

My first direct commit :tada:

sam-github:
Wicked, I just bumped into the `# SKIP undefined` this morning... I won't miss it when its gone. :-)

PR #125, opened 2015-05-12
omit tests and examples from published tarball

rmg:
The tests include many examples of working and broken TAP reports that
were not present in previous versions, which is unfriendly to CI jobs
that scan aggressively for *_/_.tap test reports.

@isaacs this is assuming the empty `.npmignore` that was already in the repo was there intentionally.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,3 @@
+test
+test-disabled
+coverage-example


isaacs:
Cool, yeah, that stuff shouldn't even be in the git repo, because it's out of date and pointless.

PR #127, opened 2015-05-12
File globbing

### LINKED ISSUE #126: npm@1 doesn't support file globbing on windows
Looks like npm@1 no longer runs the files list through glob, and Windows doesn't provide file globbing in the shell, so setting your npm test script to `tap test/test-*.js` is no longer cross-platform :-(

rmg:
Fixes #126 

@isaacs PTAL

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -272,6 +272,15 @@ else
 if (bail)
   process.env.TAP_BAIL = '1'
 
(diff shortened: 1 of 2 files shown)


isaacs:
Thanks!  That was definitely an oversight.  Landed and pushed to 1.0.3.

PR #135, opened 2015-05-23
upgrade signal-exit and nyc dependencies

bcoe:
This pull in the new versions of signal-exit and nyc, see:

https://github.com/bcoe/signal-exit/pull/9


isaacs:
Travis failure due to 0.8.  Will fix and then merge.

PR #136, opened 2015-05-26
mocha: handle error arg in done(err)

bajtos:
Allow individual tests to fail the test by passing an error to the `done` callback provided to `it` and `describe` methods.

An example of a typical Mocha tests that uses this feature:

``` js
it('reads file contents', function(done) {
  fs.readFile('foo.txt', function(err, text) {
    if (err) return done(err); // <-- THIS IS THE IMPORTANT PART
    expect(text).to.contain('foo');
    done();
  });
});
```

R=@isaacs ?

### DIFF
--- a/lib/mocha.js
+++ b/lib/mocha.js
@@ -19,8 +19,9 @@ function it (name, fn) {
 
   if (fn && fn.length)
     c.test(name, { at: at }, function (tt) {
-      fn(function () {
-        tt.end()
+      fn(function (err) {
+        if (err) tt.threw(err)
+        else tt.end()
       })
     })
   else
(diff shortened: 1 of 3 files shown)


bajtos:
Note: while I have added a test, I haven't seen it fail because the test suite checks only the first ~47 lines of the output and ignores the rest, at least when run on my machine (Node v0.10, io.js 2.1). I don't have time to investigate this issue, it seems like a bug in [test/test.js](https://github.com/isaacs/node-tap/blob/8865b966c01fded8fa64bc266754b018c6766eec/test/test.js) to me.

bajtos:
Also FWIW, unit-tests on Node v0.10 are failing on the master branch too, I believe the problem is not caused by changes introduced in my patch.

isaacs:
No, the regression on 0.10 is some unfortunate interactions between the `signal-exit` and `exit-code` modules, which both do some ugly monkey-patching to work around the shoddy exit setting/inspection capabilities in Node.  I'll try to work that out asap, and then land this after.

isaacs:
Landed on 306925dfb1980698df400db7f9c18b5b39a134a3.  Thanks!

PR #141, opened 2015-06-10
tap: suppress TAP header in subtests

rmg:
There appears to be no official standard for subtests, but the tap4j
that the Jenkins TAP plugin uses seems to expect only one header and
barfs if the subtests include their own headers.

See http://instanttap.appspot.com/

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -407,6 +407,8 @@ function childStream (self, child) {
         return
       if (line.match(/^Bail out!/))
         bailedOut = true
+      if (line.match(/^\s*TAP version \d+$/))
+        return
       if (line.trim())
         line = '    ' + line
       self.push(line + '\n')
--- a/test/test/bail-fail-spawn-bail.tap
+++ b/test/test/bail-fail-spawn-bail.tap
@@ -1,7 +1,6 @@
 TAP version 13
     # Subtest: bail fail
         # Subtest: ./test/test/nesting.js
-        TAP version 13
             # Subtest: nesting
             1..2
                 # Subtest: first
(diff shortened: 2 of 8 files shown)


rmg:
Fixed the tests and commit message, includes regenerated fixtures. Shall I merge once Travis is green? Or do you want to mull it over a bit more?

isaacs:
Discussion in npm irc channel: https://gist.github.com/isaacs/1a31aecd161de6ab8da6

isaacs:
Travis failures are because the save-file test is taking enough ms to be rounded to `1s` in the timing output.

Worth fixing, but unrelated to this PR.  LGTM!

rmg:
@isaacs I can merge but publishing is still up to you. Would you mind cutting a release?

isaacs:
Re release:

``````
22:54 <@isaacs> rmg: So, I am going to replace t.tearDown(fn)  (and
                also t.beforeEach, t.beforeAll, t.afterAll,
                t.afterEach)
22:55 <@isaacs> rmg: it's almost done but there's an issue I need to
                work out re the root test.  see the `teardown`
                branch on isaacs/node-tap
22:57 <@isaacs> rmg: since that's almost done, and these things are
                subtle behavior changes anyway, maybe we could push
                off until a 1.3 release?
22:57 <@isaacs> I don't think these behavior changes are *breaking*,
                per se, since all the user-facing stuff stays
                consistent.
22:58 <@isaacs> but you could argue that "now with tap4j support!"
                is a new feature of sorts```
``````

rmg:
I've got some config changes I'd like to roll out to StrongLoop's CI that are waiting on this patch so that it doesn't result in a flood of passing tests failing at the last minute when parsing the TAP results. (Yay Jenkins!)

PR #142, opened 2015-06-10
Make less hay

rmg:
I'm not sure whether we want the inner, the outer, or both, but I'm assuming we don't want all the levels in between?

Here are 2 different degrees of filtering for the repeated bailout messages.

### DIFF
--- a/test/test/spawn-bail.tap
+++ b/test/test/spawn-bail.tap
@@ -15,8 +15,7 @@ TAP version 13
               ---
               {"found":1,"wanted":"1","compare":"===","at":{"file":"test/test/spawn.js","line":22,"column":10,"function":"foo"},"source":"tt.equal(1, '1', 'nested failure')\n"}
               ...
-            Bail out! # nested failure
-        Bail out! # nested failure
+
     Bail out! # nested failure
 Bail out! # nested failure
 
--- a/lib/test.js
+++ b/lib/test.js
@@ -405,7 +405,7 @@ function childStream (self, child) {
     lines.forEach(function (line) {
       if (bailedOut)
         return
-      if (line.match(/^Bail out!/))
+      if (line.match(/^\s*Bail out!/))
         bailedOut = true
       if (line.match(/^\s*TAP version \d+$/))
         return
--- a/test/test/bail-child-bail.tap
+++ b/test/test/bail-child-bail.tap
@@ -6,6 +6,5 @@ TAP version 13
           {"at":{"file":"test/test/bail-child.js","line":7,"column":7},"source":"t.fail('this fails')\n"}
           ...
         Bail out! # this fails
-    Bail out! # this fails
 Bail out! # this fails
 
--- a/test/test/bail-child.tap
+++ b/test/test/bail-child.tap
@@ -6,6 +6,5 @@ TAP version 13
           {"at":{"file":"test/test/bail-child.js","line":7,"column":7},"source":"t.fail('this fails')\n"}
           ...
         Bail out! # this fails
-    Bail out! # this fails
 Bail out! # this fails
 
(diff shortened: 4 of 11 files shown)


rmg:
@isaacs does this take care of the extra bailout messages you found alarming?

isaacs:
Well, the alarming thing isn't that the extra bailouts were there, but that they were _sometimes_ there.

I mostly like this, but I'd prefer to put the `bailedOut = true` below the `if (bailedOut) return` bit, so that we get the first bailout message where it happens, and then one at the top where it's parsed and relevant.

rmg:
@isaacs no problem, that's the first commit in this PR (6a52ad2).

rmg:
@isaacs dropped the second commit, shall I merge?

test/test/spawn-bail.tap
```diff
@@ -15,8 +15,7 @@ TAP version 13
               ---
               {"found":1,"wanted":"1","compare":"===","at":{"file":"test/test/spawn.js","line":22,"column":10,"function":"foo"},"source":"tt.equal(1, '1', 'nested failure')\n"}
               ...
-            Bail out! # nested failure
-        Bail out! # nested failure
+
```
isaacs, line 18: This one is weird...  Why's that extra line showing up?  Is it killing the child proc before the `Bail out!` line?

rmg, line 18: I see a blank line pushed if the child closes before it is `_ended`, and then `bailOnFail` prints `\nBail out!...\n`.. so maybe combination of those.

rmg, line 18: Actually, fiddling with `bailOnFail` yielded what looks to me like more correct output.

isaacs, line 18: It's fine either way.  My main concern is that it's consistent.

rmg, line 18: The other oddity with this line is the "inner" bailout isn't where the actual bailout comes from.

rmg, line 18: It can be cleaned up with this change.. but it seems a little mysterious to me what the _other_ effects might be that aren't covered by tests.

```
diff --git a/lib/test.js b/lib/test.js
index e74818e..3e99084 100644
--- a/lib/test.js
+++ b/lib/test.js
@@ -578,7 +578,7 @@ function bailOnFail (self, stream, parser) {

   parser.on('assert', function (res) {
     if (!res.todo && !res.skip && !res.ok)
-      stream.emit('data', '\nBail out! # ' + res.name + '\n')
+      stream.emit('data', 'Bail out! # ' + res.name + '\n')
   })
 }

diff --git a/test/test/spawn-bail.tap b/test/test/spawn-bail.tap
index 36e92ee..ae69adf 100644
--- a/test/test/spawn-bail.tap
+++ b/test/test/spawn-bail.tap
@@ -15,7 +15,6 @@ TAP version 13
               ---
               {"found":1,"wanted":"1","compare":"===","at":{"file":"test/test/spawn.js","line":22,"column":10,"function":"foo"},"source":"tt.equal(1, '1', 'nested failure')\n"}
               ...
-
-    Bail out! # nested failure
+            Bail out! # nested failure
 Bail out! # nested failure
```

isaacs, line 18: When I re-generate the fixture, it puts it back where it is expected to be... so... yeah, there's definitely still some kind of race there.

rmg, line 18: I found v0.12.4 and iojs produced different results.. v0.12.4 produced the smallest diff, so that was what I've used for regenerating the fixtures.

rmg:
Looks like the core of this PR was landed in d1646ece30f3869d201a1afbd00ce0f3cf98ce4a

test/test/spawn-bail.tap
isaacs, line 18: I just ran the re-gen 2 times and got different results.  This is not version-specific.  Definitely a race.

rmg, line 18: :cry:

PR #143, opened 2015-06-10
Teardown

isaacs:
Adds `tap.tearDown(fn)`.

Note the long commit message in a2ad291.  There's a bit of weirdness in this change.

Still needs tests.  Please don't land yet.

7fb7094 is a bit unrelated, probably should port that over to master and remove it from this branch.

r? @rmg @bcoe


lib/root.js
```diff
@@ -15,6 +15,14 @@ process.on('exit', function (code) {
     process.exit(1)
 })
 
+// need to autoend if a teardown is added.
+// otherwise we may never see process.on('exit')!
+tap.tearDown = tap.afterAll = function (fn) {
+  var ret = Test.prototype.afterAll.apply(this, arguments)
+  this.autoend()
```
rmg, line 22: It seems odd to me that `tap.tearDown()` != `t.tearDown()`... if they have to be different (and I think that may be the case), does it even make sense to have a root level `tap.tearDown()`?

isaacs, line 22: Root-level `tap.tearDown()` is a must-have unfortunately.  The lack of it is keeping many would-be tap users on 0.x

isaacs, line 22: I guess it's not _that_ weird, since tap.test !== t.test, but maybe it ought to be pre-bound like those other methods?

lib/test.js
```diff
@@ -88,6 +89,20 @@ function Test (options) {
   Readable.apply(this, options)
 }
 
+Test.prototype.beforeEach = function (fn) {
+  this.on('beforechild', fn)
```
rmg, line 93: Is this useful when it only supports synchronous operations?

lib/root.js
rmg, line 22: Maybe make `tap.tearDown` root-only method and have `Test.prototype.afterAll` not aliased as `.tearDown` so the same method on instances and pseudo-instances of `Test` are consistent in behaviour?

lib/test.js
isaacs, line 93: I did a bit more discussion with @bcoe about the use cases for mocha and lab tests.

The simple answer is "no, not at all".  The longer answer is that we probably need to either do this in a completely different way, or just leave it out of node-tap and move it into some other module with the explicit aim to re-implement the mocha BDD API in terms of tap.

rmg, line 93: That's kinda what I thought. I think it would be reasonable to do it outside of node-tap.

rmg:
@isaacs @bcoe @sam-github since I wasn't there, was there any discussions/conclusions on this from nodeconf?

bcoe:
:+1: I will make an effort to give this a thorough review tonight.

rmg:
I'm selfishly more interested in when the next release can be cut :-P

lib/test.js
bcoe, line 93: I think that having `mocha`ish methods that don't behave identically to their namesakes introduces a bit of confusion. I'm a big fan of gradually splitting this functionality into its own module that has feature parity with mocha.

lib/test.js
```diff
@@ -88,6 +89,20 @@ function Test (options) {
...
+
+Test.prototype.afterEach = function (fn) {
+  this.on('afterchild', fn)
+}
+
+Test.prototype.afterAll = function (fn) {
+  this.on('end', fn)
+}
+
+Test.prototype.tearDown = Test.prototype.afterAll
```
bcoe, line 104: Could `tearDown` accept an optional `done` method?

``` js
// wait for done.
tap.tearDown(function (done) {
  setTimeout(function () {
    return done()
  }, 1000)
})

// autoend
tap.tearDown(function () {
  setTimeout(function () {
    console.log('never do this')
  }, 5000)
})
```
- if `done` is provided, test doesn't end and wait for for it to be executed.
- if no `done` parameter is expected in the closure provided, fallback to the `autoend` logic? -- would this potentially make it easier to implement mocha-style hooks?

isaacs:
@bcoe I'd rather not add any sort of "fork behavior based on callback arity" behavior.  Mocha does this, and I kind of hate it.

I think it's a pretty good idea to scrap after/before each, and _just_ add tearDown on the root object.  On that root object, autoend is pretty low-hazard, and so far in using this in projects that use tearDown, it's always a case of actually _wanting_ autoend to work more or less like it's implemented here.

All the before/after each stuff needs to be thought through more carefully, I think, and the mochalike behavior really needs to be a separate module.

I'm going to close this PR and open a new one to only add tearDown to the root test.

PR #157, opened 2015-07-29
Handle circular references in errors. Fixes #156

### LINKED ISSUE #156: t.error fails on circular references
Test case:

``` js
var cycle = { };

cycle.foo = cycle;

test('foo', function(t) {
  t.ifError(cycle, 'foo');
  t.end();
});
```

This fails with a `RangeError: Maximum call stack size exceeded`.

A less-obviously-made-to-fail case where this happens is if the argument to `ifError` is a response object that contains circular references.

---

The culprit is the [cleanYamlObj](https://github.com/isaacs/node-tap/blob/master/lib/test.js#L876-912) method that recurses down the object.

tmcw:
This modifies the cleanYamlObj method to handle circular references
in data, replacing the reference with [Circular] instead of
throwing an error.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -873,7 +873,7 @@ Test.prototype.printResult = function printResult (ok, message, extra) {
   this._maybeAutoend()
 }
 
-function cleanYamlObj(object, isRoot) {
+function cleanYamlObj(object, isRoot, seen) {
   if (object === undefined)
     return null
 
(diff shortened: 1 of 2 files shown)


tmcw:
Tests fail on 0.8 in Travis, not sure why. They pass on every other node version.

rmg:
The failure on 0.8 seems to exist on master as well.

isaacs:
Landed on master, after fixing the 0.8 problems.

Thanks!

PR #162, opened 2015-08-21
Add --harmony_arrow_functions CLI flag

ghost:
I like to use arrow functions, even if they are not fully implemented, the basic syntax is there. I tried the following unsuccessfully, hence this PR.

```
node --harmony --harmony_arrow_functions ./node_modules/tap/bin/run.js test.js
```


ghost:
I found it easier to run tap/tape using `babel-node`, so this is not really useful anymore. :smile:

PR #163, opened 2015-08-27
Fail on unfinished tests

### LINKED ISSUE #161: tap runs tests from `process.on('exit'`
Tests should not be run from `process.on('exit', ...`, because any async activity that is scheduled will not actually occur.

I doubt this is intentional, it may just be a side effect of https://github.com/isaacs/node-tap/issues/152, but I am reporting seperately because it is not immediately obvious why a test that is clearly scheduling an async activity is allowing node to exit.

Example:

```
'use strict';

var tap = require('tap');

tap.test('t1', function(t) {
  //t.plan(1);
  process.on('some event that never happens', function() {
    t.pass('ok');
  });
});

tap.test('t2', function(t) {
  console.log('t2 running after node is exiting? %j', process._exiting);
  process.nextTick(function() {
    // Never runs, wtf?
    t.end();
  });
});

```

Output:

```
TAP version 13
    # Subtest: t1
    1..0
ok 1 - t1 # time=12.3ms

    # Subtest: t2
t2 running after node is exiting? true
```

In above,
- test `t1` is passing wrongly, it was never ended
- test `t2` is running in the exit event handler... so it fails to actually run its next tick and pass (its a test that can never fail)
- the exit status is `0`, wrongly, the last test never ended and has no `ok`

isaacs:
Fix #161

This is a pretty significant change.  It may introduce some annoying
test failures in some cases, but most likely, those are tests that were
silently being ignored before.

Now, when `t.endAll()` is called, if there's anything waiting in the
queue, it will be treated as a failure, and reported with extreme
prejudice.  It should be virtually impossible after this change to ever
have a test simply be ignored or have the tap stream cut off after
starting a new child test.

r? @sam-github @rmg

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -137,16 +137,18 @@ Test.prototype.setTimeout = function (n) {
 Test.prototype._onTimeout = function () {
   var s = this
   while (s._currentChild && (s._currentChild instanceof Test)) {
+    s._queue = []
+    s.end()
     s = s._currentChild
   }
 
   // anything that was pending will have to wait.
-  s._queue = []
   s.fail('timeout!', {
     expired: this._name,
     timeout: this._timeout,
     at: this._calledAt
   })
+  s.end()
 
   this.endAll()
 }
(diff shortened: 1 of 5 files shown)


lib/test.js
```diff
@@ -157,15 +159,55 @@ function domainError (er) {
...
 Test.prototype.endAll = function () {
   var child = this._currentChild
+
+  if (this._queue && this._queue.length)
+    this._queueFail()
+
   if (child) {
     child._queue = []
-    if (!child._ended && !child._plan && child.fail) {
+    if (!child._ended && -1 === child._plan && child.fail)
```
isaacs, line 207: Note the bug here: `!child._plan` isn't the right test.  Before the child has a plan, it's `t._plan` is set to -1!

Issue #129, opened 2015-05-13
behaviour changes in tap@1


rmg:
Some breaking changes are expected, but now that I've converted (or attempted to convert) a handful of small and large modules, here are some notes.
#### sensitivity to debug messages

The new architecture combines stdout from tests with TAP output, so `console.log()` in tests can easily corrupt the test results. This is particularly troublesome when upgrading tests that make liberal use of the previous behaviour of auto-commenting stdout from tests and is the main thing blocking me from upgrading some modules.
#### only first child test is run if tests aren't ended

Another change is that nested tests that contain only nested tests are no longer ended when the grandchildren have all ended. Ultimately this is probably more of a correctness improvement, so it might make sense to just add some sort of detection so people's previously passing tests don't just fail with `not ok missing plan`.

The following passes under tap@0.7.1:

``` js
var tap = require('tap')

var grandChildren = []

tap.test('parent', function(t) {
  t.test('child 1', function(t) {
    t.test('grandchild 1', function(t) {
      grandChildren.push('grandchild 1')
      t.ok(1, 'child 1')
      t.end()
    })
  })
  t.test('child 2', function(t) {
    t.test('grandchild 2', function(t) {
      grandChildren.push('grandchild 2')
      t.ok(1, 'child 2')
      t.end()
    })
  })
})

tap.test('grandchildren', function(t) {
  t.equal(grandChildren[0], 'grandchild 1', 'grandchild 1 is run')
  t.equal(grandChildren[1], 'grandchild 2', 'grandchild 2 is run')
  t.end()
})
```

The above can be made to pass under tap@1 by adding a `t.end()` at the end of the `child 1` test.

@isaacs I didn't want to bother opening a PR to add the above failing test because I haven't looked yet at if/how it might be fixed.

isaacs:
> sensitivity to debug messages

Yeah, this is actually one of the things that has been a common complaint from mocha & lab users; they can't use `console.log` for arbitrary debugging, because TAP turns it into a comment, which is then ignored by the reporter.

Perhaps it'd be better to send `'extra'` events to stderr rather than passing up the chain?

> only first child test is run if tests aren't ended

I'm surprised that ever worked!  It definitely shouldn't!

[...]

PR #164, opened 2015-08-27
Fail on unfinished tests

### LINKED ISSUE #152: tests wrongly "pass" when node exits on empty event loop
This test should not pass:

```

test('Test container', function(t) {
  t.test('set env', function(tt) {
  });

  t.test('set deployment id', function(tt) {
    tt.end();
  });
});
```

It does:

```
sn/strong-central (ws-robustness *+) % node _.js     
TAP version 13
    # Subtest: Test container
        # Subtest: set env
        1..0
    ok 1 - set env # time=8.235ms

    1..1
ok 1 - Test container # time=23.56ms

1..1
# time=49.764ms
```

The "set env" test was never ended, and "set deployment id" was never even run!

The test can be made to detect that it has failed by adding a `{timeout: 2000}` options object to the top-level test.

@rmg did some poking around, and it seems that the runner's timeout is achieved by wrapping the spawn in a timeout. I don't think this is sufficient, it appears to me that test objects, at least the top-level "default" test object, should have a default timeout (one that is overridden by TAP_TIMEOUT, if the runner has set that env variable).

This might be related to https://github.com/isaacs/node-tap/commit/22680b0e15428fff218e12615d4f10d4a28b1546#commitcomment-12328000, wherein test are also not running, and yet the test is "passing".

### LINKED ISSUE #161: tap runs tests from `process.on('exit'`

isaacs:
Fix #161
Fix #152
Re #129

This is a pretty significant change.  It may introduce some annoying
test failures in some cases, but most likely, those are tests that were
silently being ignored before.

Now, when `t.endAll()` is called, if there's anything waiting in the
queue, it will be treated as a failure, and reported with extreme
prejudice.  It should be virtually impossible after this change to ever
have a test simply be ignored or have the tap stream cut off after
starting a new child test.

r? @sam-github @rmg

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -337,6 +337,8 @@ var TMR = require('tap-mocha-reporter')
 if (files.length === 1 && files[0] === '-') {
   // if we didn't specify any files, then just passthrough
   // to the reporter, so we don't get '/dev/stdin' in the suite list.
+  // We have to pause() before piping to switch streams2 into old-mode
+  process.stdin.pause()
   reporter = new TMR(reporter)
   process.stdin.pipe(reporter)
   process.stdin.resume()
--- a/lib/test.js
+++ b/lib/test.js
@@ -137,16 +137,18 @@ Test.prototype.setTimeout = function (n) {
 Test.prototype._onTimeout = function () {
   var s = this
   while (s._currentChild && (s._currentChild instanceof Test)) {
+    s._queue = []
+    s.end()
     s = s._currentChild
   }
 
   // anything that was pending will have to wait.
-  s._queue = []
   s.fail('timeout!', {
     expired: this._name,
     timeout: this._timeout,
     at: this._calledAt
   })
+  s.end()
 
   this.endAll()
 }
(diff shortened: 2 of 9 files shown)


isaacs:
io.js failure is spurious: https://github.com/nodejs/node/issues/2581

Other failure was legit.  This patch exposed a test that wasn't ending itself properly!  The system works!

rmg:
Couple thoughts..
- test/test/unfinished.js could really use some comments describing what is happening, like which lines are run and which are never run
- would it be possible for the "XXX left in queue" messages to include the test name if found? Might be more human friendly than only providing the details in the yaml block.

sam-github:
Thanks @isaacs, should be much more robust.

PR #165, opened 2015-08-27
deepEqual: don't blow up when comparing circular objects

othiym23:
Uses the newly-created [`only-shallow`](http://npm.im/only-shallow) to handle `deepEqual`'s duck-typed structural equality testing, and [`deeper`](http://npm.im/deeper) to handle the stricter equality checking required by `strictDeepEqual`. `only-shallow` was built specifically for this use case, but it's worth reading the source to make sure that what it does agrees with your understanding of `deepEqual`'s semantics.

Since `deeper` and `only-shallow` can both handle Buffer testing, there's no need to depend on `buffer-equal`. `deeper` has an optional dependency on @bnoordhuis's `buffer-tools`, which may or may not be faster than using a tight loop in JavaScript these days, but thanks to the way that `deeper` uses it, it shouldn't prevent browserification or do anything else bad. If it turns out to be a problem in practice, I'm not really sure the dependency is worth the trouble anymore, so I could be persuaded to remove it.


rmg:
I'm having difficulty navigating Travis, but I _think_ the failure on 0.10 is on master, but the failure on iojs is introduced by this PR somehow.

othiym23:
@rmg It's possible, but #164 is failing on io.js in [exactly the same way](https://travis-ci.org/isaacs/node-tap/jobs/77604003). I'm not really sure how my changes could cause a V8 failure on their own, unless `buffertools` was in the mix somehow.

othiym23:
@rmg As a followup, I just verified that the failures with 0.10 are also failures on master. If I get some time this weekend, I'll look into figuring out what's up with both.

isaacs:
io.js introduced a regression in 3.0 that breaks tap's tests. https://github.com/nodejs/node/issues/2581

When Travis upgraded to 3.3, it got annoying for us here.  I'm gonna take a look at see how to lock travis down to io.js 2.x.

isaacs:
Hmm... buffertools makes an annoying time.  Is it possible to get a deeper that doesn't use that?  On io.js it results in an attempt to download an invalid tgz from nodejs.org, all kinds of weirds happening which scare users, even though it ends up working.

othiym23:
@isaacs: sure, give me a minute. As I said above, it's possible to document that it will take advantage of `buffertools` if it's available, because it already falls back to a JS implementation if the optional dependency fails.

othiym23:
@isaacs Rebased against the new master, and included a version of `deeper` with no dependencies!

rmg:
Green CI: :white_check_mark: 
Net-negative diffstat: :white_check_mark: 
Net-positive test count: :white_check_mark: 

:shipit:

isaacs:
Thanks!

PR #169, opened 2015-09-03
Add support for running tests via babel, with source-mapped stack traces

myndzi:
I don't know if this is something you'll want or not, but it's here if you do. I had a choice of shimming it at `Test#spawn` or in `run.js`, but chose the latter since the former is an external api and would require me to parse arguments to the node command line in a way I'm not sure would be very robust.

### DIFF
--- a/lib/babel-wrap.js
+++ b/lib/babel-wrap.js
@@ -1,0 +1,15 @@
+var path = require('path')
+
+// remove ourselves from process.argv
+var idx = process.argv.indexOf(__filename)
+if (idx > -1) process.argv.splice(idx, 1)
+
+// install source maps so that stack dumps will be correct
+require('source-map-support').install()
+
+// install babel
+require('babel/register')
+
+// require the original target file, export its exports
+var fullpath = path.join(process.cwd(), process.argv[1])
+module.exports = require(fullpath)
--- a/lib/babel-wrap.js
+++ b/lib/babel-wrap.js
@@ -1,8 +1,7 @@
 var path = require('path')
 
 // remove ourselves from process.argv
-var idx = process.argv.indexOf(__filename)
-if (idx > -1) process.argv.splice(idx, 1)
+process.argv.splice(1, 1)
 
 // install source maps so that stack dumps will be correct
 require('source-map-support').install()
--- a/bin/run.js
+++ b/bin/run.js
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 var args = process.argv.slice(2)
 
+var path = require('path')
 var fs = require('fs')
 if (!args.length && process.stdin.isTTY) {
   console.error(usage())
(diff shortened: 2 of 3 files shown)


lib/babel-wrap.js
```diff
@@ -0,0 +1,16 @@
+var path = require('path')
+
+// remove ourselves from process.argv
+process.argv = process.argv.filter(function (arg) {
```
isaacs: This filter function is going to result in some false positives.  Probably it's better to just splice out the exact filename that we know it's going to be?

myndzi: You're right; this was a leftover from trying to support it as a relative path, which was unfortunate to begin with.

myndzi:
Updated it to just remove index 1, which is always "this file". In general I can't imagine why a test would require the property that argv[1] === __filename, but I wanted the wrapper to be as transparent as possible. You'll notice that the wrapper is still referenced in error information dumped by tap -- this is a side-effect of inserting this shim in run.js instead of the Test class, but I don't see it as a problem.

myndzi:
It seems that this approach doesn't entirely work yet -- I have an instance of a test that is failing because it expects another assertion; whatever is generating the details when printing the TAP output is pointing to the wrong place when using the es6 source :(

Edit: looks like writeDiags is directly reading the file and accessing the lines from there. Would have to build in source map support to test.js for this to work out...

isaacs:
I wonder if it would be better to use spawn-wrap to automagically babelify everything (ie run the scripts with babel-node) and then just have tap be blissfully unaware. 

Http://npm.im/spawn-wrap

myndzi:
The trouble is that you need extra help for sourcemaps server side, so wrapping in babel-node is not enough. Since this is about testing, having line numbers that don't match with the pre-translated source is not helpful at all. That's why the wrapper here also loads `source-map-support`, but unfortunately this doesn't work with tap's line sourcing in some cases, such as incomplete tests which are supposed to map to the line where the test is declared.

Making wrapping more generic would certainly be helpful to people who write in, say, coffeescript, but solving the stack trace / information aspect is equally important. My guess is that anything that can output a source map will work with the `source-map-support` module, and perhaps with a little work, the Test class can be made to check for and interpret source maps if they exist. That's a bit beyond my familiarity with the existing code, for now, though.

myndzi:
[ This wrapping process also seems to mess up coverage... :( ]

myndzi:
Too many problems with this approach, I think. I'm gonna deal with it in my fork and maybe I'll get somewhere helpful. I shouldn't have PR'd from master so I need to close this, but feel free to comment if you want me to create another PR or something.

isaacs:
Thanks, @myndzi.  I'm looking forward to what you come up with.  Feel free to reach out with any questions or bounce ideas here or elsewhere.

A pretty valid approach would be to just do nothing and wait for Node to include all of Babel's ES6 stuff natively ;)

myndzi:
That timing...

isaacs:
I know right? I honestly didn't yet know node 4 had just been released when I wrote that comment ;)

Issue #171, opened 2015-09-16
Issue when using promises


louisameline:
Hello, I'm seeing a weird behavior when using promises.

This dummy example works as expected : 

```
tap.test('test', function(t){

    $.Deferred().resolve()
        .then(function(){
            var def = $.Deferred();
            setTimeout(def.resolve, 1000);
            return def.promise();
        })
        .then(function(){
            t.ok(true);
        })
        .always(function(){
            t.end();
        });
});
```

while my real-life example fails :

```
tap.test('kvstore', function(t){

    $.Deferred().resolve()
        .then(function(){
            console.log('A');
            return KVStore.connect();
        })
        .then(function(){
            console.log('B');
            t.ok(true);
        })
        .always(function(){
            console.log('C');
            t.end();
        });
});
```

This is the output : 

```
A
B
C
test/lib/KVStore.test.js .............................. 1/3 30s
  not ok timeout
  not ok missing plan
    results:
      ok: false
      count: 2
      pass: 1
      fail: 1
      failures:
        - ok: false
          id: 2
          name: timeout
          parent: ''

total ................................................. 1/3


  1 passing (30s)
  2 failing

```

Note: my KVStore.connect currently returns a promise on the connection to a postgres database. Using `t.plan(1)` does not work either.

There are no other tests scheduled before or after this one. Using Node 0.12 and Tap 1.4.1. Thanks

rmg:
What does `KVStore.connect()` do? If it opens a connection, that will keep the event loop alive. Then, even if your tests complete, the process won't exit and will timeout instead.

louisameline:
Oh, I wouldn't have guessed that. Thank you !

rmg:
@isaacs is there a way to list active handles in the timeout failure message that doesn't rely on the undocumented methods in core? Might be useful.. I know I often end up adding it in as debug output when trying to figure out why a process isn't exiting when I think it should..

isaacs:
Since the timeout happens at the parent process, and the handles are in the child process, no, it would be impossible.

One option might be to listen on `SIGTERM` and emit a failure in those cases?  It still won't help with child procs that aren't running node-tap, but oh well.

rmg:
Ah, right.

PR #172, opened 2015-09-17
Emit a more useful failure on child proc timeout

isaacs:
Catch the SIGTERM and output something intelligible about what
the currently active handles/requests are.

Inspired by #171 

Review, please: @rmg @louisameline


lib/root.js
```diff
@@ -29,3 +29,46 @@ tap.mocha = require('./mocha.js')
 tap.mochaGlobals = tap.mocha.global
 tap.Test = Test
 tap.synonyms = require('./synonyms.js')
+
+// Handle sigterm signals, since this usually indicates a timeout
+process.on('SIGTERM', function onTerm () {
+  process.removeListener('SIGTERM', onTerm)
+  if (process.listeners('SIGTERM').length === 0) {
```
rmg: What does this conditional do if tests set their own `SIGTERM` listener for some reason?

test/test/pending-handles.tap
```diff
@@ -0,0 +1,19 @@
+TAP version 13
+    # Subtest: ___/.*/~~~pending-handles.js child
+    ok 1 - this is ok
+    not ok 2 - received SIGTERM with pending event queue activity
+      ---
+      {"handles":{"0":{"msecs":100000,"type":"Timer"},"1":{"connectionKey":"6::::8000","events":{"0":"request","1":"connection","2":"clientError"},"type":"Server"},"2":{"events":{"0":"end","1":"finish","2":"_socketEnd","3":"free","4":"close","5":"agentRemove","6":"drain","7":"error","8":"data"},"type":"Socket"},"3":{"events":{"0":"end","1":"finish","2":"_socketEnd","3":"drain","4":"timeout","5":"error","6":"close","7":"data","8":"resume","9":"pause"},"type":"Socket"}},"requests":{"0":{"type":"FSReqWrap"}}}
```
rmg: Should the handles and requests not be expanded out to yamlish?

louisameline:
Yes that's much clearer, thank you very much.

lib/root.js
isaacs: It would remove itself and do nothing else.

test/test/pending-handles.tap
isaacs: The output here is turned into JSON for the test fixtures just so that it's easier to evaluate when running the tests if for example the key order differs or some other irrelevant detail.

Technically JSON is a subset of YAML, so it's also valid TAP, but yeah, it won't look like that in real life :)

isaacs: Oh, in case I misunderstood your question: I experimented with just dumping the request objects as-is to the YAML encoder, but you end up with a big glut of function strings and mostly-useless nonsense.  The goal of this output should be to give you a clue as to what might be going on, and if we just dump everything, there's a high likelihood of drowning out the useful bits in a bunch of noise.

isaacs:
The downside of this, unfortunately, is that signals are a bit of a fiddly beast.

On Travis, about half the time the output fails to be flushed to stdout before the fatal self-SIGTERM hits.  I'm looking into a way to make sure it waits for the drain, but it might require some more clever workarounds.

isaacs:
AH!  no, the problem is that `signal-exit` is being used by the coverage stuff, and that adds another signal handler.  So, it works fine as long as you're not using coverage ;)

Back to the drawing board.

isaacs:
I got around the multiple-listener thing easily enough using `signal-exit` rather than `process.on('SIGTERM')`.  However, there's still a race condition between receiving the SIGTERM, emitting the failure output, and then triggering the self-inflicted fatal SIGTERM.  I'm a bit uncomfortable just pushing the timeout values out further and further until it works for me, especially since travis-ci and other systems are going to have different timing characteristics.  When coverage is enabled, everything gets a lot slower, so the timeouts have to be longer, and if the timeouts are too long, then the user starts to feel the impact.

test/test/pending-handles.tap
rmg: I had a feeling it was fixture-specific formatting. And I agree, just dumping the full handle would be so noisy it would become useless.

isaacs:
Dug more into this.  The issue is process startup time.  When nyc loads istanbul, it takes about an extra 30ms of sys time, and 400ms of wall-clock time.  Probably I just need to make the tests take longer when coverage is enabled, since there are likely not a lot of people writing tests that time out after 300ms anyway.

isaacs:
Ok, this is in a pretty reasonable place to be merged.  I think it's a net improvement, but wow, what a long and weird road.

In the process, I learned that (a) most of the `test/test/*.tap` files weren't actually being parsed properly by `test/test.js`. Whenever it entered a yaml block, it'd never leave that yaml block, and then the file would end, and it'd be like, "Welp, no errors, guess it passes!", so only the lines before the first yaml block were ever tested!

I fixed that issue, and (b) to my dismay _every test/test.js fixture failed_, because at some point in the move to having `t.match` and `t.has` the same function, and having it handle regexps, and then switching from `deep-equal` to using `only-shallow` and `deeper`, we dropped the regexp support for parsing strings that live inside objects.

So, I created a new module, `tmatch`, which does deep "pattern matching".  That is, if you have an object like: `{ a: { likely: /ring$/ } }`, then that'll "match" the object `{extra: 1, a: { likely: 'String' }}`.  Also, this means that `t.has` is a `t.has` all the way down; before, it would select the top-level fields, but anything below that level was tested with deep equivalence.

Of course, this is a breaking change.  If anyone was banking on the subtle difference between `t.match(obj,pattern)` _not_ supporting regexps, then their tests might start passing incorrectly now.  On the other hand, it's closer to how it was documented, so... I'm torn.

Probably this much hand-wringing means it's time for a bump to `tap@2.x` here.

isaacs:
and this is the type of output you can expect when handles are left open causing a timeout:

```
$ tap never.js
never.js .............................................. 0/1 405ms
  ./never.js child
  not ok received SIGTERM with pending event queue activity
    requests:
      - type: FSReqWrap
    handles:
      - type: Timer
        msecs: 100000
      - type: Server
        events:
          - request
          - connection
          - clientError
        connectionKey: '6::::8000'
      - type: Socket
        events:
          - end
          - finish
          - _socketEnd
          - free
          - close
          - agentRemove
          - drain
          - error
          - data
      - type: Socket
        events:
          - end
          - finish
          - _socketEnd
          - drain
          - timeout
          - error
          - close
          - data
          - resume
          - pause

total ................................................. 0/1


  0 passing (424.708ms)
  1 failing
```

or in raw tap:

``` tap
TAP version 13
    # Subtest: never.js
        # Subtest: ./never.js child
        not ok 1 - received SIGTERM with pending event queue activity
          ---
          requests:
            - type: FSReqWrap
          handles:
            - type: Timer
              msecs: 100000
            - type: Server
              events:
                - request
                - connection
                - clientError
              connectionKey: '6::::8000'
            - type: Socket
              events:
                - end
                - finish
                - _socketEnd
                - free
                - close
                - agentRemove
                - drain
                - error
                - data
            - type: Socket
              events:
                - end
                - finish
                - _socketEnd
                - drain
                - timeout
                - error
                - close
                - data
                - resume
                - pause
          ...
        1..1
        # failed 1 of 1 tests
        # time=81.715ms
    not ok 1 - ./never.js child # time=227.643ms
      ---
      timeout: 200
      at:
        file: never.js
        line: 20
        column: 3
      failure: timeout
      results:
        pass: 0
        plan:
          start: 1
          end: 1
        ok: false
        count: 1
        fail: 1
      signal: SIGTERM
      command: /usr/local/bin/node
      arguments:
        - /Users/isaacs/dev/js/tap/never.js
        - child
      source: |
        t.spawn(process.execPath, [__filename, 'child'], {}, '', { timeout: 200 })
      ...

    1..1
    # failed 1 of 1 tests
    # time=250.977ms
not ok 1 - never.js # time=396.223ms
  ---
  arguments:
    - never.js
  results:
    count: 1
    fail: 1
    pass: 0
    ok: false
    plan:
      start: 1
      end: 1
  timeout: 30000
  exitCode: 1
  command: /usr/local/bin/node
  file: never.js
  ...

1..1
# failed 1 of 1 tests
# time=414.217ms
```

othiym23:
LGTM. I like that the code is more explicit now about which of `shallower`, `deeper`, or `tmatch` is being used in any given case. The new behavior around signals also seems like a welcome change. Might cause some carnage in `npm`, but for the better, and so worthwhile. :sheep:!

PR #173, opened 2015-09-29
Remove domains

isaacs:
Fixes a problem where expected EventEmitter-related throws are
impossible to test, because the domain hijacks the 'error' event.

Of course, this means that asynchronous unexpected throws will always be
assigned to the global root TAP object, but... what can you do?  If
people complain, we can try to figure out something clever using
continuation-local-storage or something.

Ugh.  Domains.  The not-quite-worst-possible solution to any problem,
which always ends up regrettably.

I am, as always, truly and deeply sorry for inflicting this feature on
Node.js

### DIFF
--- a/lib/root.js
+++ b/lib/root.js
@@ -30,6 +30,14 @@ tap.mochaGlobals = tap.mocha.global
 tap.Test = Test
 tap.synonyms = require('./synonyms.js')
 
+process.on('uncaughtException', function onUncaught (er) {
+  var child = tap
+  while (child._currentChild && child._currentChild instanceof Test) {
+    child = child._currentChild
+  }
+  child.threw(er)
+})
+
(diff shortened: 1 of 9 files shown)


rmg:
Apology in the form of removing tap's use of domains accepted.

As for using CLS... please make it opt-in if you do use it. Maybe it's just bad luck, but CLS seems to offer stacktrace obfuscation as a bonus feature, which isn't really what you want when debugging a failing test.

isaacs:
> As for using CLS... please make it opt-in if you do use it. Maybe it's just bad luck, but CLS seems to offer stacktrace obfuscation as a bonus feature, which isn't really what you want when debugging a failing test.

Good to know!  Mostly, I just think that the async throw tracking is kind of a shame to lose.  It still tracks the thrown error, of course, just catches it at the top level instead of closer to its origin.

PR #174, opened 2015-10-06
Use promises to end tests or fail them

hybrist:
I tested this manually but I'm struggling with writing the proper tests. Don't have time right now to dig into how they work, so if anyone would be interested in picking this up...


hybrist:
Wow, that was quick - thanks a lot! :)

isaacs:
I fixed up a few things to make it produce more helpful output.

For future reference, the way to generate the .tap test fixtures is like this:

```
node scripts/generate-test-test.js test/test/*.js
```

I should make it an npm script or document that, probably.

PR #175, opened 2015-10-07
Allow cleanup code to run in process.exit

hybrist:
While trying to add `tap`-support for our [integration test tool](https://github.com/groupon/testium) we ran into problems with leaking processes. It comes down to the following pseudo-code:

``` js
var tap = require('tap')

// ... somewhere in another module
var child = spawn('phanomjs')
process.on('exit', function() { child.kill() })

// back in the test code
tap.test('some failing test', function(t) {
  t.equal(1, 2)
  t.end() // oops, `child` will never exit
})
```

I'm not sure how to solve this problem for node 0.10 and before but starting with 0.12 `tap` could use `process.exitCode` instead of calling `process.exit` which lets other `process#exit` listeners run and (for us) fixes everything.

### DIFF
--- a/lib/root.js
+++ b/lib/root.js
@@ -16,8 +16,16 @@ tap.on('bailout', function () {
 process.on('exit', function (code) {
   tap.endAll()
 
-  if (!tap._ok && code === 0)
-    process.exit(1)
+  if (!tap._ok && code === 0) {
+    // `process.exitCode` was added in node 0.12.
+    // https://github.com/nodejs/node/commit/a5dba82ee210685ffce17b35ff2da9cca1d268d2
+    if (/^v(0\.12|[1-9])/.test(process.version)) {
+      // Allow cleanup code in other process#exit handlers to run.
+      process.exitCode = 1
+    } else {
+      process.exit(1)
+    }
+  }
 })
 
 // need to autoend if a teardown is added.
--- a/test/exit-code.js
+++ b/test/exit-code.js
@@ -6,7 +6,7 @@ var path = require('path')
 var fixtures = path.resolve(__dirname, 'fixtures')
 
 test('exit code 1 when tap results show failure', function (t) {
-  t.plan(3)
+  t.plan(4)
 
   t.test('test exits 0, has failures', function (t) {
     t.plan(2)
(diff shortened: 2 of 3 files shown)


rmg:
@jkrems unrelated to the actual PR, but for the problem you are describing...

One approach that works well with node is instead of killing the child processes is opening an IPC channel as part of the spawn and then in the child detecting when that pipe goes away and using that as a signal to exit. It's not directly applicable since you are spawning a process that is out of your control, but maybe it gives you some ideas for approaching the problem from another direction.

hybrist:
@rmg I could in theory try to spawn one more level of processes (since I neither control the application under test nor phantomjs) that handles the killing. But I'm hesitant to add another level of complexity. Once I go for hacks, I'd rather monkey-patch `process.exit` to do the cleanup instead or in addition to using a `process#exit` handler... ;)

P.S.: Sorry, forgot to say thanks for the suggestion. Appreciate the help. :) Shouldn't write comments while in meetings.

rmg:
@jkrems ya, I sympathize. I was hoping there was something specific to your case that would allow you to side-step the wrapper process approach.. I wonder if phantomjs would accept a PR that adds a parent IPC check/monitor.

hybrist:
Yeah, child processes are depressingly hard to get right. We also fork the app under test in our tool. Requiring any app that wants to use it to add such a check would be a bit awkward. So, even when we get it into phantomjs, it only gets us half the way there. :(

isaacs:
When the parent dies, unless the child is detached, it should receive a SIGHUP signal, which is normally fatal.  It could be that phantomjs is catching and ignoring that SIGHUP, since you're not running it detached.

We can use `process.exitCode`, but there are cases where we really want to make sure that we exit, but iirc, there are cases where `exitCode` changing in the `exit` event isn't caught properly, which is why this was changed initially.

It'd probably be safest to just make sure that we're always calling `process.exit()` in the last `on('exit')` handler.  I'm not sure if that's possible without clubbing the `process.emit` function.

More research is needed.  In the short term, you can set up your `process.on('exit')` handler ahead of loading `tap`, which should cause the behavior you want.

hybrist:
@isaacs So the default behavior of node could be considered a bug?

``` bash
node -e "console.log(require('child_process').spawn('node', [ '-e', 'setTimeout(function(){}, 60000)' ]).pid); process.exit()"
```

After that, the pid of the child is still alive (and will be until it exits due to the empty event loop a minute later).

hybrist:
Found a solution in `subprocess` that works without changes to `tap`. Thanks for looking into this! :)

PR #176, opened 2015-10-07
Protect against out-of-bound line numbers

hybrist:
When using require hooks (e.g. `babel-node`), the stack positions of a test failure do not necessarily resolve to a valid line in the file on disk. This change gracefully ignores them. A better fix might be to get source mappings for babel/coffee/etc. working but this prevents crashes at least.


rmg:
I love these kinds of little changes.. an obvious one-liner fix with a proportional boat load of tests :+1: 

Landed in 46915c526bdb5403389f70e1aef222d1784093f9 (with a couple minor tweaks).

PR #177, opened 2015-10-16
Process queue on _afterEnd event, not 'end'

### LINKED ISSUE #153: end event for a test occurs after the next test has already started
Basically with

```
test('a', function() {
  t.on('end', END
}
test('b', function() {
   .... this test will start before the END function of test a runs...
```

This is not so useful. I'm using t.on(end) because when tests end I need to some cleanup before the next test. Since the test is quite async, and uses a plan, I don't really know when everything is done, not without reimplementing the plan...

I can dump cleanup for test a at the beginning of test b, and I'm about to do that, but I think that this behaviour is not right.

Runnable reproduction in test form:

```
var tap = require('tap');

var run = null;
// timeout to work around #152
tap.test('parent', {timeout: 1000}, function(t) {
  tap.test('a', function(t) {
    console.assert(run === null); run = 'a';

    t.plan(1);

    t.on('end', function() {
      console.error('END: a == %j', run);
      console.assert(run === 'a'); run = null;
    });

    setImmediate(function() { t.assert(true); });
  });

  tap.test('b', function(t) {
    console.assert(run === null); run = 'b';

    t.plan(1);

    t.on('end', function() {
      console.error('END: b == %j', run);
      console.assert(run === 'b'); run = null;
    });

    setImmediate(function() { t.assert(true); });
  });

  t.end();
});
```

output:

```
% TAP_BAIL=1 node _.js                                                                                                                                    
TAP version 13
    # Subtest: parent
    1..0
ok 1 - parent # time=10.589ms

    # Subtest: a
    1..1
    ok 1 - expect truthy value
ok 2 - a # time=10.239ms

    # Subtest: b
    not ok 1 - AssertionError: false == true
      ---
      at:
        file: console.js
        line: 87
        column: 23
        function: Console.assert
      name: AssertionError
      actual: false
      expected: true
      operator: ==
      generatedMessage: true
      test: b
      message: 'AssertionError: false == true'
      stack: |
        Console.assert (console.js:87:23)
        _.js:20:13
        Test.test (node_modules/tap/lib/test.js:346:27)
        Test._processQueue (node_modules/tap/lib/test.js:716:13)
        Test.<anonymous> (node_modules/tap/lib/test.js:336:10)
        Test.end (node_modules/tap/lib/test.js:701:10)
      ...
    Bail out! # AssertionError: false == true
Bail out! # AssertionError: false == true
END: a == "a"
```

isaacs:
This should fix https://github.com/isaacs/node-tap/issues/153

@sam-github would you be so kind as to review and make sure that it addresses your problem?

Thanks!

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -386,6 +386,8 @@ Test.prototype.test = function test (name, extra, cb) {
     self.ok(child._ok, name, extra)
     if (!self._ended)
       self.push('\n')
+  })
+  child.on('_afterEnd', function () {
     self._processQueue()
   })
   child.on('bailout', function (message) {
@@ -777,9 +779,11 @@ Test.prototype.done = Test.prototype.end = function end () {
 
   this.emit('complete', final)
   var self = this
-  if (!this._endEmitted)
+  if (!this._endEmitted) {
+    this._endEmitted = true
     this.emit('end')
-  this._endEmitted = true
+    this.emit('_afterEnd')
+  }
 }
 
 Test.prototype._processQueue = function () {
--- a/lib/test.js
+++ b/lib/test.js
@@ -387,9 +387,6 @@ Test.prototype.test = function test (name, extra, cb) {
     if (!self._ended)
       self.push('\n')
   })
-  child.on('_afterEnd', function () {
-    self._processQueue()
-  })
   child.on('bailout', function (message) {
     rootBail(self, message)
   })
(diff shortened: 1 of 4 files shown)


rmg:
Is there any concern that relying on the event handler being synchronous turns this into more of a hook point than a notification? Not sure if that's a distinction anyone actually cares about..

Does seem to address @sam-github's issue, though, as far as I can tell.

isaacs:
Yeah, actually, that shouldn't even be an "event", just a method or something.

PR #180, opened 2015-10-20
Add automatic Codecov.io support

evanlucas:
Add support for piping coverage results to codecov.io via the
CODECOV_TOKEN environment variable.

Ref: https://github.com/isaacs/node-tap/issues/179

It does not include any tests as I wasn't sure how we wanted to go about testing the code coverage. Any input on that would be greatly appreciated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "node": ">=0.8"
   },
   "dependencies": {
+    "codecov.io": "0.1.6",
     "coveralls": "^2.11.2",
     "deeper": "^2.1.0",
     "foreground-child": "^1.2.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -11,6 +11,7 @@
     "node": ">=0.8"
   },
   "dependencies": {
+    "codecov.io": "0.1.6",
```
rmg, line 14: Is it a good idea to use an exact version here? It seems to be updated several times a year..

evanlucas, line 14: Definitely yalls call on that. I always thought that for 0.x versions, semver would not actually satisfy versions other than the one specified

isaacs:
Landed in 688776107f7dd0c2a92c98822809612c9b164761

Thanks!

PR #182, opened 2015-10-25
initial implementation for piping coverage data to multiple services

dotnetCarpenter:
I need some help to test this. Basically I'm pretty blank when it comes to how I would test this. I can insert environment variables but how do I spy on _coveralls.js_ and _codecov.io.js_ too see how there are called?

I'll take deeper look into how the other tests are made in _runner.js_ and update this PR but help is wanted ;)

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -59,8 +59,6 @@ var coverage = pipeToService
 var coverageReport
 
 var nycBin = require.resolve('nyc/bin/nyc.js')
-var coverallsBin = require.resolve('coveralls/bin/coveralls.js')
-var codecovBin = require.resolve('codecov.io/bin/codecov.io.js')
 
 for (var i = 0; i < args.length; i++) {
   var arg = args[i]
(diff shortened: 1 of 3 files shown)


dotnetCarpenter:
I tried to set an environment variable (in test/runner.js) for the test case and output some data in the export logic (in bin/run.js) but since exporting coverage data is spawned in another process, I had a lot of timing issues and in the end gave up.

I'm now thinking to refactor out the export logic to a separate and testable module.

dotnetCarpenter:
Would be good to have some [spy](https://github.com/philbooth/spooks.js) functionality in `tap` for this.

dotnetCarpenter:
I've spend some time thinking about how I can write proper tests for this. IPC and signals are new to me so I've done a lot of reading and I have some ideas. But first, I would like to check my assumptions.
- [x] this patch is working as is
- [x] I understand how the current test process is working
- [ ] ~~'ipc' or~~ signals can be used to communicate between test and `run.js` (but carries an significant overhead due to spawning too many sub-processes)
- [ ] a module using EventEmitter can manage processes while giving feedback to a `require` consumer

dotnetCarpenter:
**this patch is working as is**

node-tap-coverage-pr

See https://travis-ci.org/Firefund/firefund-cli/jobs/88444497
Based on: #183

dotnetCarpenter:
**I understand how the current test process is working**
node-tap test process
https://www.lucidchart.com/invitations/accept/34b4fb46-7152-456f-9b83-2e140f28919c

dotnetCarpenter:
Confirmed that **I understand how the current test process is working**

See https://github.com/dotnetCarpenter/node-tap/blob/pipecover-understanding/understand.js

node-tap-understand

dotnetCarpenter:
Hmm.. Since the PR is working locally but not on travis, I'm at a loss for how to debug this. I do get some failing tests but they are in `runner.js`(1), `test.js`(4) and `pending-handles.js`(1). When I print out data I also get a fail due to how `tap` reads tests from stdout.

I've tried to use 'ipc' to communicate between run.js (the coverage export logic) and test to no avail. process,listeners('messages') is always empty in `run.js`, no matter what I do in a test.

Since this (signals) also failed, I'm out of ideas. The patch is working fine #183 but I can't figure out how to test it. which is a shame. Unless you have any ideas or pointers to how this can be tested - I'm giving up :-1:

isaacs:
This is addressed in v3.1.

Issue #181, opened 2015-10-25
Pipe coverage to both Coveralls AND Codecov


dotnetCarpenter:
> If a COVERALLS_REPO_TOKEN environment variable is set, then coverage is
> captured by default and sent to the coveralls.io service. If a CODECOV_TOKEN
> environment variable is set, then coverage is captured by default and sent to the
> codecov.io service.

I am checking out both [Coveralls](https://coveralls.io/) and [Codecov](https://codecov.io/) for the first time and it's pretty easy with `node-tap`. Unfortunately not at the same time(!)

If COVERALLS_REPO_TOKEN is set in travis then `tap` only pipe to Coveralls regardless that CODECOV_TOKEN is also set. The order doesn't matter, coverage data is only piped to Codecov if there is no COVERALLS_REPO_TOKEN variable.

dotnetCarpenter:
I think line 236-256 in run.js should be re-factored to pipe to both services. https://github.com/isaacs/node-tap/blob/master/bin/run.js#L236

dotnetCarpenter:
I gave it a shot with PR #182

isaacs:
Landed in 3.1

PR #183, opened 2015-10-30
added fix for #181 but without any changes to test

dotnetCarpenter:
Hi @isaacs 
Concluding that my first assumption is correct. **Patch #182 is working as is**

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -59,8 +59,6 @@ var coverage = pipeToService
 var coverageReport
 
 var nycBin = require.resolve('nyc/bin/nyc.js')
-var coverallsBin = require.resolve('coveralls/bin/coveralls.js')
-var codecovBin = require.resolve('codecov.io/bin/codecov.io.js')
 
 for (var i = 0; i < args.length; i++) {
   var arg = args[i]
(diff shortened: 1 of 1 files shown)


isaacs:
Hey, @dotnetCarpenter, sorry for the delay on this.  I haven't forgotten about it, just haven't found time to review carefully.  Since it's doing some interesting juggling with streams and child procs, I want to make sure I understand all the moving parts before merging, even if it seems to be working properly.

dotnetCarpenter:
@isaacs in short; only the _services_ array is new and I wrapped [line 244-260 in @evanlucas commit in a forEach](https://github.com/isaacs/node-tap/commit/688776107f7dd0c2a92c98822809612c9b164761#diff-1316baf4c35bbaca22d1c8c3ab02bb59R244) and deleted some stuff that is now redundant or not needed.

bin/run.js
```diff
@@ -59,8 +59,6 @@ var coverage = pipeToService
 var coverageReport
 
 var nycBin = require.resolve('nyc/bin/nyc.js')
-var coverallsBin = require.resolve('coveralls/bin/coveralls.js')
```
dotnetCarpenter, line 62: redundant, since the introduction of the _services_ array

bin/run.js
```diff
@@ -228,39 +226,45 @@ if (coverageReport && !global.__coverage__ && files.length === 0) {
   var args = [nycBin, 'report', '--reporter', coverageReport]
   var child
 
-  // automatically hook into coveralls
+  // automatically hook into coveralls and/or codecov
   if (coverageReport === 'text-lcov' && pipeToService) {
     child = spawn(node, args)
-    var covBin, covName
```
dotnetCarpenter, line 234: not needed since we have the _services_ array

bin/run.js
```diff
@@ -228,39 +226,45 @@ if (coverageReport && !global.__coverage__ && files.length === 0) {
...
-
-    if (process.env.COVERALLS_REPO_TOKEN) {
```
dotnetCarpenter, line 236: the main issue of #181 that #183 is fixing. Here we say if else if, in #183 we say forEach

bin/run.js
```diff
@@ -228,39 +226,45 @@ if (coverageReport && !global.__coverage__ && files.length === 0) {
...
-      covBin = coverallsBin
-      covName = 'Coveralls'
-    } else if (process.env.CODECOV_TOKEN) {
-      covBin = codecovBin
-      covName = 'Codecov'
-    }
 
-    var ca = spawn(node, [covBin], {
```
dotnetCarpenter, line 244: moved inside the forEach

bin/run.js
```diff
@@ -228,39 +226,45 @@ if (coverageReport && !global.__coverage__ && files.length === 0) {
...
+      , covName: 'Coveralls'
+      }
+    , process.env.CODECOV_TOKEN && {
+        covBin: require.resolve('codecov.io/bin/codecov.io.js')
+      , covName: 'Codecov'
+      }
+    ].filter(function(s) {
+      return !!s // remove undefined services
     })
-    signalExit(function (code, signal) {
```
dotnetCarpenter, line 257: moved inside the forEach

bin/run.js
```diff
@@ -228,39 +226,45 @@ if (coverageReport && !global.__coverage__ && files.length === 0) {
...
-      child.kill('SIGHUP')
-      ca.kill('SIGHUP')
+
+    services.forEach(function(s) {
```
dotnetCarpenter, line 246: forEach is new

bin/run.js
```diff
@@ -228,39 +226,45 @@ if (coverageReport && !global.__coverage__ && files.length === 0) {
...
+      var ca = spawn(node, [s.covBin], {
```
dotnetCarpenter, line 247: The content is pretty much the same as it has always been. Since the forEach argument is a function, we get a function scope environment. That is, each variable is stored in the environment which in turns is stored on the heap.
Given this code, we spawn a child process and assign it to **ca** in this environment. We call the _child._**stdout.pipe** function with our instance of **ca.stdin**, basically telling the _nyc/bin/nyc.js_ process to pipe its output to our **ca.stdin**. This happens for each service (maximum 2, minimum 0), since _child._**stdout.pipe** can pipe to several functions, this is a non issue.
We then attach an _on_ handler that is stored in the _child_process.spawn_ function scope environment. Lastly, we call **signalExit** with a new function that has access to **child** and **ca** in this environment via a pointer/reference.
If we have 2 services, then we have 2 environments which each spawn a child process in parallel (_var ca = spawn(node, [s.covBin], {..._). We create new function scope environments for each of these processes, so there can be no conflict.

The only thing that can lead to concern is the question of when the _signalExit_ function argument is run. If we got 2 environments and both have access to **child** and they both call **child.kill('SIGHUP')**, is the code then racy? I haven't seen it but I also don't know the internals of [signal-exit](https://github.com/bcoe/signal-exit).

See the _"Environments: Managing Variables"_ section of [Speaking JavaScript](http://speakingjs.com/es5/ch16.html) for an explanation of javascript environment.

bin/run.js
```diff
@@ -228,39 +226,45 @@ if (coverageReport && !global.__coverage__ && files.length === 0) {
...
+      })
+      signalExit(function (code, signal) {
+        child.kill('SIGHUP')
+        ca.kill('SIGHUP')
+      })
     })
   } else {
     // otherwise just run the reporter
-    var child = fg(node, args)
+    child = fg(node, args)
```
dotnetCarpenter, line 267: why `var` when we do it on [line 227](https://github.com/isaacs/node-tap/pull/183/files#diff-1316baf4c35bbaca22d1c8c3ab02bb59R227)

dotnetCarpenter:
@isaacs I hope I answered your question with my last [(updated) comment](https://github.com/isaacs/node-tap/pull/183/files#diff-1316baf4c35bbaca22d1c8c3ab02bb59R247). If not, then please ask again, so I can clarify.

isaacs:
Finally got around to landing this, with some updates so that it would work when running tap's tests with coverage turned on.  (The double-wrap weirdness was making it fail strangely.)

PR #185, opened 2015-11-11
make t.throws match if regex is passed

evanlucas:
The documentation states that "The expected error is tested against the throw error using `t.match`, so regular expressions and the like are fine.", but passing a regex as the second argument does not actual verify that it matches.

The following currently passes:

```
t.throws(function() {
  throw new Error('1')
}, /test/)
```

This change makes sure that the follow would instead fail as documented.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -213,8 +213,12 @@ function decorate (t) {
       // 'name' is a getter.
       if (er.name)
         er.name = er.name + ''
-      if (wanted)
+      if (wanted) {
+        if (Object.prototype.toString.call(wanted) === '[object RegExp]') {
+          return this.match(er.message, wanted, m, e)
+        }
         return this.has(er, wanted, m, e)
+      }
       else
         return this.pass(m, e)
     }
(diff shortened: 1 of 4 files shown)


rmg:
I noticed this the other day and I'm glad someone beat me to it.

rmg:
I wonder is this should actually be considered a bug in tmatch and fixed there. @isaacs what was your intention for this behavior? You clearly meant for it to be implemented somewhere, but node-tap and tmatch each seem to assume the other covers it.

isaacs:
First of all, good catch, clear patch, thorough test, awesome job @evanlucas.  Thanks!

Re whether this should be here or in tmatch, I think it might be weird to treat an `Error` object as a string sometimes and an object other times.  For example, `t.throws(..., { message: /^ENOENT/, syscall: 'open' })` would do exactly what you would expect today.  I like just special-casing this in `t.throws` since that's a bit of an oddball anyway.

isaacs:
Landed on 2.2.1.  Thanks!

Issue #187, opened 2015-11-14
Exclusive test, .only method


dy:
Hi!
Is there a way to run any test exclusively, like [mocha](http://mochajs.org/#exclusive-tests)’s `it.only` or `describe.only`?
I guessed something like

``` js
t.test.only(function() {});
```

But there seems no that option.
That feature is quite useful in mocha, what is supposed way to quickly isolate single test in tap’s workflow?

sindresorhus:
:+1: Should support multiple `test.only()` tests in contrast to Mocha which only allows one.

isaacs:
This is tricky in node-tap.  Here's why:

``` javascript
var t = require('tap')
console.log('before')
t.pass('this is ok')
t.test('child', function (t) {
  t.plan(1)
  t.pass('child is ok')
})
console.log('mid')
t.test('child 2', function (t) {
  t.plan(1)
  t.pass('child 2 is ok')
})
console.log('after')
```

The output is this:

```
$ node t.js
before
TAP version 13
ok 1 - this is ok
    # Subtest: child
    1..1
    ok 1 - child is ok
ok 2 - child # time=6.092ms

mid
    # Subtest: child 2
    1..1
    ok 1 - child 2 is ok
ok 3 - child 2 # time=4.846ms

after
1..3
# time=41.233ms
```

In other words, `t.test` just runs right away, before moving on, so it can't know if there is another test coming up in the future with an `only` flag.  The reason for this is that I want `console.log()` calls to be in the "expected" place so it's easier to tie the output to the js test.

rmg:
What _would_ be possible (maybe even trivial? famous last words..) to implement would be `t.skip()`..

Completely untested, but I'm imagining something like this:

``` js
Test.prototype.skip = function skip (name, extra, cb) { // matching .test()
  // some minimal effort to get a useful name
  if (typeof name === 'function') {
    name = name.name;
  }
  name = name || 'test'; 
  // this is basically what `{skip: true}` does when passed as the second arg to `test()`
  return this.pass('skipped: ' + name);
}
```

isaacs:
@rmg I'd approach it like this:

``` JavaScript
Test.prototype.skip = function (name, extra, cb) {
  if (typeof name === 'function') {
    cb = name
    name = name.name || ''
    extra = {}
  } else if (typeof extra === 'function') {
    cb = extra
    extra = {}
  }

  if (!extra)
    extra = {}

  extra.skip = extra.skip || true
[...]

PR #189, opened 2015-11-18
Use function.name if name arg isn't provided

isaacs:
This supports stuff like:

``` javascript
t.test(function someTestName (t) { ... })
```

I noticed some other test frameworks do this, and I think it's a cute API.  Super trivial to implement.


rmg:
Funny, when I used it in my example in #187 I thought this was already supported.

isaacs:
Ok, I'mma merge then.  SemVer-minor bump coming shortly.

PR #191, opened 2015-12-15
fix name style to reflect actual name in code

### LINKED ISSUE #190: t.teardown is not a function
In the [README, it says that t.teardown](https://github.com/isaacs/node-tap#tteardownfunction) exist but it is `undefined`.

tap: 2.3.1

Example:

``` js
test('setting port via cli - default port', function(t) {
  t.plan(2);

  var port = defaultPort;
  var options = ['.'];
  var ecstatic = startEcstatic(options)

  console.log(t.teardown); // <- output undefined

  t.teardown(function() {
    ecstatic.kill('SIGTERM');    
  });
});
```

dotnetCarpenter:
Calling it literally `teardown()` when it is actually `tearDown()` confused me and probably others. Fixes #190

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,7 +301,7 @@ by using [Travis-CI](https://travis-ci.org) and
 The root `tap` object is an instance of the Test class with a few
 slight modifications.
 
-1. The `teardown()`, `plan()`, and `test()` methods are pre-bound onto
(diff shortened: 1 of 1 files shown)


isaacs:
rad, thanks!  I also aliased it to `teardown` in case anyone has out of date docs or something :)

PR #193, opened 2015-12-17
Add --flags option

### LINKED ISSUE #160: Feature request: pass flags to io.js / node
io.js allows flags to be passed in on startup, of which the most interesting one for me is `-r <package>`.
With `tape` it's possible to do `node -r 'babel/register' test`, it'd be nice if there was a similar way to do that with `tap`.

E.g. maybe something like:

```
tap --flags '-r \'babel/register\'' **/test.js
```

Is this something that's desirable?

**edit**: Alternatively the `-r, --require` flag could be added but that seems less flexible / future proof than having direct access to the node process.

tmcw:
This implements @yoshuawuyts's suggestion and fixes #160 by adding a `--flags` value to tap.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -164,6 +164,11 @@ for (var i = 0; i < args.length; i++) {
       nodeArgs.push('--harmony')
       continue
 
+    case '--flags':
+      val = val || args[++i]
+      nodeArgs = nodeArgs.concat(val.split(' '))
+      continue
+
     case '--color':
       color = true
       continue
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
neat!

isaacs:
Work is underway on this in the `flags-n-stuff` branch. https://github.com/tapjs/node-tap/tree/flags-n-stuff

I need to land some changes in nyc to make it testable, and then bring the coverage back up to where it should be.

isaacs:
Did this in a different way, as `--node-arg=--blerg` which can be specified multiple times to provide multiple args, like `tap --node-arg=--strict --node-arg=--expose_gc --node-arg=-r --node-arg=babel`

PR #197, opened 2015-12-24
bump nyc

jamestalmage:
The only breaking change is the switch to glob patterns for excluding files.

Improvements include:
- Better handling of require hooks.
- Better support for Babel (and other transpilers) with  `--require babel-core/register`.
- Lots of Windows fixes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "glob": "^6.0.1",
     "js-yaml": "^3.3.1",
     "mkdirp": "^0.5.0",
-    "nyc": "^3.1.0",
+    "nyc": "^5.0.1",
     "only-shallow": "^1.0.2",
     "opener": "^1.4.1",
     "readable-stream": "^2.0.2",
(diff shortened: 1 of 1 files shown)


jamestalmage:
The one failing test is isolated to Node 4 (which is where I thoroughly tested before pushing) and looks to be timing related. That particular VM ran much slower than the others, so I suspect a fluke. I am about to push with an amended commit message to re-trigger. 

[Link to failed build](https://travis-ci.org/isaacs/node-tap/builds/98635909) - I suspect it was a fluke.

jamestalmage:
Yep. Looks like a fluke, second build is already past that point without a failure: https://travis-ci.org/isaacs/node-tap/jobs/98636760

jamestalmage:
:tada:

isaacs:
Thanks!  Yes, the glob/regex config switch is what I wasn't getting.  Thanks for the patch :)

jamestalmage:
@isaacs - I saw you cut a patch release with this. I think this also represents a breaking change for `tap`. Specifically for users configuring `nyc.excludes` as regular expressions in `package.json` and using the `--coverage` flag. I should have made that clearer in the PR, sorry.

isaacs:
Huh.  Nothing in tap's documentation or API directly exposes that, but you're right, since users can interact with nyc via their package.json, it's sort of global.

I'll unpublish the patch and re-publish as a major version bump.

PR #200, opened 2015-12-30
install standard to fix builds

fscherwi:


isaacs:
Whoops, thanks for that.  Forgot to add to package.json.

PR #201, opened 2015-12-31
Update appveyor.yml

alexindigo:
Should fix the builds.
re: http://www.appveyor.com/docs/lang/nodejs-iojs

PS. Also you can add `io.js` with

```
    - nodejs_version: '1.0'
```

PPS. Not sure that node@0.10 would work with latest npm, unless there is more magic that I'm aware of. :)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,7 @@ environment:
     - nodejs_version: '0.12'
     - nodejs_version: '0.10'
 install:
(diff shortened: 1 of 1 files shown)


alexindigo:
Nope, didn't fix. :) But at least now it's failing during tests :)

isaacs:
Yeah, I updated this myself and tracked down the failure to #202.

alexindigo:
:+1:

PR #206, opened 2016-01-02
Travis CI also build on osx

fscherwi:
Also run Travis build on osx to run all major platforms!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: false
-osx:
+os:
   - linux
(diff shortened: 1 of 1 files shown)


isaacs:
Thanks, but since most of the devs working on node-tap already use OS X as their main machines, it gets plenty of coverage in that OS, and this lib doesn't really have any linux- or bsd-specific bits, so all unixes are more or less the same.  I'd rather not make the travis build take longer, or rely on nvm.

PR #213, opened 2016-01-07
replace `lib/stack` with external module.

### LINKED ISSUE #207: Split `./lib/stack.js` into it's own module.
There's plenty of goodness in there I'd love to port to AVA.

I have already extracted it here: https://github.com/jamestalmage/stack-utils, and written a quite a few tests around it.

The API is very similar, except the export is a constructor that allows you to customize `cwd` and the list of `internals` regular expressions.

If you are willing, I can submit a PR that uses that module.

### LINKED ISSUE #211: stack.js TypeError: Cannot read property 'replace' of undefined
I'm reading a file in my test suite that may or may not exist.

When it doesn't exist, I get this error:

``` js
Error
    at Error (native)
    at Object.fs.openSync (fs.js:549:18)
    at Object.fs.readFileSync (fs.js:397:15)
... and some more lines
```

The important bit is `at Error (native)` - which makes stack.js `parseLine` fail because it expects every error line to have a file. [here is the offending line in stack.js](https://github.com/tapjs/node-tap/blob/5183ec8ee8536fbad6f2f3edb8774bd8b456ecfa/lib/stack.js#L218) that assumes `file` (`match[8]`) is defined. In this case it is not; `line.match(re)` produces this output for me:

``` js
[
  '    at Error (native)', // 0
  undefined, // 1
  'Error', // 2
  undefined, // 3
  undefined, // 4
  undefined, // 5
  undefined, // 6
  undefined, // 7
  undefined, // 8
  undefined, // 9
  undefined, // 10
  'native', // 11
  index: 0, // 12
  input: '    at Error (native)' // 13
]
```

jamestalmage:
`stack-utils` also contains a fix for the `at Error (native)` problem (See #211).

Fixes #207
Fixes #211


jamestalmage:
That second commit should almost certainly be a separate PR. Just trying to hunt down why the first commit failed (passes locally on Windows).

jamestalmage:
OK. Looks like AppVeyor is just a timing failure. Almost certainly unrelated to this. Oddly the first commit shows as passing now. I did amend that commit due to a linter error I missed, maybe that caused the `red X` I was seeing next to that first commit until just recently.

@isaacs do you want me to drop that second commit?

isaacs:
Thanks, I landed this on master, published as 5.0.1, and tweaked the appveyor a bit myself.

PR #217, opened 2016-01-16
use `clean-yaml-object`

jamestalmage:
The `cleanYamlObj` function has been extracted to an external module.

Feedback welcome on the new API. Especially the signature of the filter function. It feels a bit bloated currently.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -21,6 +21,7 @@ var Parser = require('tap-parser')
 var path = require('path')
 var Module = require('module').Module
 var fs = require('fs')
+var cleanYamlObject = require('clean-yaml-object')
 var binpath = path.resolve(__dirname, '../bin')
 
 function hasOwn (obj, key) {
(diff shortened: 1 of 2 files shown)


bcoe:
@jamestalmage mind rebasing with my branch (the other open pull request) it looks like Windows tests finally work with @isaacs' most recent fixes to foreground-child. hooray \o/

jamestalmage:
rebased. passed locally. Let's see what AppVeyor has to say.

jamestalmage:
I'm not sure why AppVeyor failed. I think it's unrelated to this PR

isaacs:
Yeah, it's timing out.  I restarted it, should go through fine.

jamestalmage:
Node 4 and 5 passed. Node 0.12 failed with a "missing plan" failure.

 I wonder if you're running into the same issues I had with AVA's AppVeyor test suite. We would frequently see the last line or two of a child_process's output truncated for no reason. This was especially problematic on Node `0.10`, but I've seen it in all Node versions. 

I was never able to reproduce locally (I might have once, but it was like one out of 100 times instead of one out of 6 times I was seeing on AppVeyor).

 We solved the problem in AVA by rewriting our reporters to be testable without inspecting stdout/stderr (our reporters return strings, and we have a wrapper class that simply dumps those strings to stderr/stdout).  I'm not sure that's a viable strategy for `node-tap`.

isaacs:
I have zero idea why 0.12 is borking on appveyor.  Works fine for me locally.

On the change itself, I think it's a good thing to abstract out, but something feels a bit off in the interface to me.

If the goal is to clean up an object such that it can be passed to yaml, then that's fine.  But since you still have to pass in a custom filter function for tap's purposes, maybe there's an opportunity to do something a bit _more_ tap-specific and abstract that away as well?

Is ava going to use this as well?  Does it also need to pass in its own filter function?

jamestalmage:
>  something feels a bit off in the interface to me.

Yeah, I don't love it either (a big reason for the pre `1.0.0` version).

> Is ava going to use this as well? 

Yes.

> Does it also need to pass in its own filter function?

I think so (I haven't implemented yet). I am certain AVA will want to handle the `stack` property differently. Not sure about `at`.  Once I actually implement for AVA, I may find that we can standardize further and abstract out even more shared code.

I did consider making `stack-utils` a dependency and passing through `stack-utils` options. That didn't seem right either, as it reduces it from a "prepare **any** object for serialization" utility, to a "prepare **errors** for serialization" one.

If you've got some good ideas for improving it, I am all ears. Otherwise my vote would be to merge it (knowing it is in beta), and see what happens when I integrate with AVA (I've got a few other issues on my plate ahead of that). I promise to follow up with PR's here if we end up making breaking changes. I'd just as soon avoid letting the extracted code get out of sync with what you have currently.

jamestalmage:
:tada:

PR #218, opened 2016-01-16
bump stack-utils to 0.3.0

jamestalmage:
stack-utils now normalizes to unix path separators _before_ appling the `internals` filter. Regexps no longer need to account for backslashes.

stack-utils@0.3.0 also brings long-stack-trace support

https://github.com/tapjs/stack-utils/releases/tag/v0.3.0

### DIFF
--- a/lib/stack.js
+++ b/lib/stack.js
@@ -2,6 +2,6 @@ var StackUtils = require('stack-utils')
 
 module.exports = new StackUtils({
   internals: StackUtils.nodeInternals().concat([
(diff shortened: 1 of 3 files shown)


isaacs:
Appveyor failure is fixed by https://github.com/bcoe/nyc/pull/138, caused by bin/nyc.js wrapping and parsing args that should be passed to the child command.

Looks good to me!

isaacs:
Oh! I was wrong, it's not fixed by bcoe/nyc#138, it's fixed by the flags-n-stuff branch on node-tap.  I'll land that soonish.

isaacs:
This is landed on master now.  Thanks!

PR #219, opened 2016-01-24
upgrade nyc dependency

bcoe:
the `master` branch of nyc has now landed the updates that had lead to a fork.


bcoe:
@isaacs looks like Windows tests are actually passing \o/

Mind restarting the one failed run, looks like an intermittent install problem which I've seen on AppVeyor before.

isaacs:
Landed on 4743640 and published as 5.1.2.  Thanks!

Issue #228, opened 2016-02-01
Use .taprc to set defaults


isaacs:
Read a config file at ~/.taprc for setting default colors, reporters, etc.

kgraves:
I'm looking into working on this, but I'm curious on how you want it done.

There are modules out there that parse rc files. Are you against using any of these?
What format do you want to support? (e.g. .ini, .json, .yml, etc)
Any other ideas on features/support that you would like to see?
Should cli args trump the .rc file or vice versa? (I'm thinking the former)

isaacs:
Command line flags always are highest priority.  (Near is a higher indicator of intent than far-away.)

Only one file should be parsed: `require('os').homedir() + '/.taprc'`  (So no need for hierarchical RC file stuff.)

I'm thinking _probably_ yaml makes the most sense?  Since tap uses yaml already and js-yaml is already a dep?  And then JSON would work as well, since it's a subset of yaml.

Keys should match the CLI args.

Args that can be set multiple times (`nyc-arg` and `node-arg`) can be specified either as a single string, or an array of strings.  (Since you can't set a key multiple times in yaml like you can on a command line.)

isaacs:
Also, thanks!  Looking forward to it :)

isaacs:
Use [os-homedir](https://www.npmjs.com/package/os-homedir) to polyfill for 0.10 and before.

kgraves:
Now that I'm working on this issue, I have a question. When it comes to rc file format (json vs yml), did you want me to support both json and yml, or just choose one over the other? I couldn't decipher this from your response :)

kgraves:
@isaacs ^ There's my first pass, let me know what you think.

isaacs:
@kgraves Json is a subset of yaml. So, use yaml, and then anyone who wants json is taken care of.

kgraves:
(facepalm) Gotcha, I probably should have thought about that a little more. Now that I reread that, it's pretty clear.

isaacs:
Landed in 6.0.0

PR #234, opened 2016-02-10
adds .taprc file support #228

kgraves:


bin/run.js
```diff
@@ -93,26 +107,37 @@ function main () {
...
   }
 
-  options.color = supportsColor
   if (process.env.TAP_COLORS !== undefined) {
-    options.color = !!(+process.env.TAP_COLORS)
+    defaultArgs.color = !!(+process.env.TAP_COLORS)
+  }
+
+  if (process.env.TAP === '1') {
+    defaultArgs.reporter = 'tap'
```
isaacs: This should actually be set on the final option object, not just the default set.

test/utils.js
```diff
@@ -0,0 +1,52 @@
...
+  })
+
+  t.test('returns empty object when not valid yaml', function (t) {
+    var expected = {
+      timeout: 9999,
+      coverage: false,
+      coverageReport: false,
+      reporter: 'classic'
+    }
+    t.strictSame(utils.parseRcFile(validYmlPath), expected)
```
isaacs: Should this be checking a file that's not valid yaml?  Maybe use `__filename` or something?

isaacs:
Landed and published on 6.0.0.  thanks!

kgraves:
woohoo :+1:

Issue #244, opened 2016-03-23
Name argument of Test.prototype.test is not optional when providing options object


maarten-t:
Calling `tap.test(options, testFunction)` will create a subtest with the name "[object Object]", instead of "(unnamed test)". The subtest ignores the supplied options object.

This behaviour surprises me because the documentation of Test.prototype.test suggests that all three arguments are optional:

> t.test([name], [options], [function])

isaacs:
Huh.  Yeah, that seems like a bug for sure.

Patch welcome!

PR #245, opened 2016-03-24
fix Test.prototype.test optional arguments (issue #244)

maarten-t:
I'm not sure if the tests that I added use the preferred strategy of testing the fixed feature. I wasn't able to find examples of similar tests. Some test documentation or guidelines would be helpful. Part of the problem is that it wasn't immediately clear to me which parts of the code were the tests being tested and which were the tests of the tests.

Note that there is a suspicious line in `Test.prototype.test`, which I didn't touch out of cautiousness.
When the test name is missing, it's being set to its fallback value `'(unnamed test)'` at [line 453](https://github.com/themaarten/node-tap/blob/master/lib/test.js#L453) after it is used at [line 453](https://github.com/themaarten/node-tap/blob/master/lib/test.js#L448). Is it supposed to be in that order? It would make sense to set the fallback value around [line 433](https://github.com/themaarten/node-tap/blob/master/lib/test.js#L433).

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -395,20 +395,38 @@ Test.prototype.test = function test (name, extra, cb, deferred) {
     clearTimeout(this._autoendTimer)
   }
 
-  if (typeof name === 'function') {
-    cb = name
+  var variableArgs = Array.prototype.slice.call(arguments, 0, 3);
+  
+  if (typeof variableArgs[0] === 'string'){
+    name = variableArgs[0]
+    variableArgs.shift()
+  }
+  else {
     name = ''
-    extra = {}
-  } else if (typeof extra === 'function') {
-    cb = extra
+  }
+
+  if (typeof variableArgs[0] === 'object'){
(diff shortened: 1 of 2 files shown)


isaacs:
Setting the fallback earlier would be fine, but there's no harm in the current order.  All that'll happen is that it'll re-enter the function with `undefined` or some other falsey value, and then fall back the second time around.

For tests, in this case it'd be sufficient to modify one of the examples in `test/test/*.js` to set an options object and no name.  Then run `npm run regen-fixtures` to update the output fixtures.

isaacs:
Test failures on Node 5 are unrelated.  The `function` value on throws changed from `Timer._onTimeout` to `null._onTimeout`, so I just need to update the test to allow for this.

isaacs:
Landed on master, but slightly different implementation, and factored it out so that this bit could be unit tested explicitly.

Thanks!

maarten-t:
Thank you!

PR #249, opened 2016-04-11
Wait for returned promises before ending tests

### LINKED ISSUE #240: Don't autoend until returned promise has fulfilled/rejected
Tests end when the planned assertion count is reached, even if a returned promise is still pending.

This is problematic when assertions are run in response to events that are triggered while the promise is pending. More events could occur but tap moves on to the next test.

Here's a reduced example:

``` js
const test = require('tap').test

test('example', t => {
  t.plan(1)

  return new Promise(resolve => {
    setTimeout(resolve, 1000)
    t.pass()
  }).then(() => console.error('fulfilled'))
})

test('after', t => {
  console.log('after')
  t.end()
})
```

Results in:

``` console
$ $(npm bin)/tap -R spec test.js

test.js
  example
    ✓ (unnamed test)

after
fulfilled
  ✓ after

  2 passing (1s)
```

But I would have expected:

``` console
$ $(npm bin)/tap -R spec test.js

test.js
  example
    ✓ (unnamed test)

fulfilled
after
  ✓ after

  2 passing (1s)
```

novemberborn:
Fixes #240.

If a promise is returned then that promise should dictate the test lifecycle. This prevents tests from ending prematurely when a planned assertion count is reached, but further work takes place before the returned promise is fulfilled (or indeed rejected).

This commit defers ending the test until the test function has returned. Care is taken to ensure that ending tests more that once, exceeding the planned assertion count, or causing exceptions after ending the test, is still reported correctly _after_ the initially deferred end has finished.

A new "promise-plan" test has been added for the scenario outlined in #240.

---

Some tests have now have extra newlines in their TAP output. This is due to https://github.com/tapjs/node-tap/blob/d5e11d0f660ac0a38b486adfb505b28a02e51fd3/lib/test.js#L543:L544 which may run when an end is deferred. I decided not to put in an extra guard since the newlines are probably harmless. I also believe this same condition may occur when using asynchronous `afterEach` callbacks, so it's not really a new issue.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -83,7 +83,8 @@ function Test (options) {
   this._currentChild = null
   this._ending = false
   this._ended = false
-  this._planFinished = false
+  this._mustDeferEnd = false
+  this._deferredEnd = null
 
   this._parent = null
   this._printedVersion = false
@@ -550,14 +551,28 @@ Test.prototype._runChild = function runChild (child, name, extra, cb) {
 
   // still need try/catch for synchronous errors
   try {
+    child._mustDeferEnd = true;
     var cbRet = cb(child)
     if (cbRet && typeof cbRet.then === 'function') {
       // promise
       cbRet.then(function () {
-        child.end(IMPLICIT)
-      }, child.threw)
+        child._mustDeferEnd = false;
+        if (!child._tryResumeEnd()) {
+          child.end(IMPLICIT)
+        }
+      }, function (err) {
+        child._mustDeferEnd = false;
+        if (!child._tryResumeEnd()) {
+          child.threw(err)
+        }
+      })
+    } else {
+      child._mustDeferEnd = false;
+      child._tryResumeEnd()
     }
   } catch (er) {
+    child._mustDeferEnd = false;
+    child._tryResumeEnd()
     child.threw(er)
   }
   self._level = self
@@ -968,6 +983,14 @@ Test.prototype.done = Test.prototype.end = function end (implicit) {
     return
   }
 
+  // Repeated end() calls should be run once any deferred ends have finished.
+  if (this._deferredEnd) {
+    this._deferredEnd.after.push(function () {
+      this.end(implicit)
+    })
+    return
+  }
+
   if (implicit !== IMPLICIT && !this._multiEndThrew) {
     if (this._explicitEnded) {
       this._multiEndThrew = true
@@ -981,6 +1004,20 @@ Test.prototype.done = Test.prototype.end = function end (implicit) {
     return
   }
 
+  // If neccessary defer the end until the child callback has returned.
+  if (this._mustDeferEnd) {
+    this._deferredEnd = {
+      implicit: implicit,
+      plan: this._plan === -1 ? this._count : this._plan,
+      after: []
+    }
+    return
+  }
+
+  this._finishEnd(implicit);
+}
+
+Test.prototype._finishEnd = function (implicit) {
   // If the afterEach function throws, then we can end up back here
   if (this._ranAfterEach || !this._parent) {
     this._end(implicit)
@@ -1088,6 +1125,23 @@ Test.prototype._end = function (implicit) {
   }
 }
 
+Test.prototype._tryResumeEnd = function () {
+  if (!this._deferredEnd) {
+    return false
+  }
+
+  var implicit = this._deferredEnd.implicit
+  var after = this._deferredEnd.after
+  this._deferredEnd = null
+
+  this._finishEnd(implicit)
+  after.forEach(function (fn) {
+    fn.call(this)
+  }, this)
+
+  return true
+}
+
 Test.prototype._processQueue = function () {
   if (this._bailedOut) {
     return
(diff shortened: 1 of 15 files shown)


test/test/console-log.tap
```diff
@@ -18,9 +18,10 @@ TAP version 13
         ok 4 - nested ok second
         1..4
     ok 2 - second ___/# time=[0-9.]+(ms)?/~~~
-ok 1 - nesting ___/# time=[0-9.]+(ms)?/~~~
 
 >>>> after second child
+ok 1 - nesting ___/# time=[0-9.]+(ms)?/~~~
+
```
isaacs, line 24: This is a relevant behavior change.  Synchronous ends should still be synchronous.

isaacs:
This is almost there.

However, tests that _don't_ return a promise (or have async `afterEach` functions) need to end synchronously.

test/test/console-log.tap
novemberborn, line 24: They are. `>>>> after second child` is logged from inside the `nesting` test. It cannot be ended after `second` completes because it may still return a promise. Note that `>>>> after child test` correctly follows `ok 1 - nesting`.

[Source](https://github.com/tapjs/node-tap/blob/2543379a8bb7eeea2161263241ada4c58c47b54d/test/test/console-log.js#L32)

isaacs, line 24: Oh, I see, so the issue here is that `nesting` was ending immediately upon the end of `nesting > second`, but now it's waiting until after the return of the `nesting` callback, because it returns a promise.

So, yeah, it's _relevant_ (perhaps major version bump?) but yes, I see now, it is correct and necessarily implied by this feature.

novemberborn:
@isaacs trying to figure out why the Node < 4 tests are failing.

test/test/console-log.tap
novemberborn, line 24: I think this could be considered a bug? Can't really see how people would be relying on this behavior.

novemberborn:
Tests should be fixed, accidentally used arrow functions 😊

novemberborn:
The Travis failure seems unrelated. AVA tests pass when I link this branch.

isaacs:
Just wanted to say that I haven't forgot about this.  I've got it working on my machine, and everything looks mostly good, it's just a big change and I haven't yet had time to fully review it.  Thanks for your patience :)

novemberborn:
@isaacs no worries 👌

lib/test.js
```diff
@@ -550,14 +551,28 @@ Test.prototype._runChild = function runChild (child, name, extra, cb) {
...
       cbRet.then(function () {
-        child.end(IMPLICIT)
-      }, child.threw)
+        child._mustDeferEnd = false;
+        if (!child._tryResumeEnd()) {
+          child.end(IMPLICIT)
+        }
+      }, function (err) {
+        child._mustDeferEnd = false;
+        if (!child._tryResumeEnd()) {
```
rmg, line 565: Shouldn't a rejection be treated the same as if an exception had been thrown?

test/test/console-log.tap
rmg, line 24: Even if they aren't intentionally relying on it, having their tests suddenly start failing without changing their own code would likely be seen as a breaking change. Given the number of major version bumps in the last 12 months I think most tap users would expect one for this change.

lib/test.js
rmg, line 565: It doesn't seems to affect the test results if this is changed to unconditionally call `child.threw(err)`. Missing test cases?

rmg, line 565: Feel free to dismiss this as a personal style preference, but I re-factored this chunk when exploring locally to look like this:

``` js
  // still need try/catch for synchronous errors
  try {
    child._mustDeferEnd = true;
    var cbRet = cb(child)
    if (cbRet && typeof cbRet.then === 'function') {
      // promise
      cbRet.then(callbackResolved, callbackThrew)
    } else {
      callbackFinished()
    }
  } catch (er) {
    callbackThrew(er)
  }
  self._level = self

  function callbackFinished() {
    child._mustDeferEnd = false
    return child._tryResumeEnd()
  }

  function callbackThrew(err) {
    callbackFinished()
    child.threw(err);
  }

  function callbackResolved() {
    if (!callbackFinished()) {
      child.end(IMPLICIT)
    }
  }
```

novemberborn, line 565: @rmg I'm not quite sure, it's been a while since I wrote the code here 😄 

If `cb(child)` throws then this code is never executed. What could happen is that the test is failed and then the promise is rejected. I suppose this would then prevent the promise rejection from being logged.

Possibly we should `_tryResumeEnd()` and regardless of its return value still run `child.threw(err)`. But it sounds like we'd need a test case for this behavior first.

isaacs:
Well... so...

3 days ago, a version of the codecov.io module was published that depended on a non-existent version of request.  In my haste to pin the dep down and prevent tap users from having failing tests everywhere (there were a lot of red travises!) I sort of accidentally published this (definitely not a patch) as 5.7.3 instead of bumping it to 5.8.0 or 6.0.0.

Anyway, it's working great and now I get to figure out if it's worth rolling out and forward-porting to a new version to make a 5.7.4 that doesn't have it, or just chalk it up to "being a human" and go on with life.

@novemberborn thanks for your patience and for this patch.  It's working great, as far as I can tell :)

@rmg I still would like to dig into some of the edge cases you brought up, but I also want to get this merged and update the docs, and (properly) call it v5.8.0.

PR #251, opened 2016-04-12
Make API link more noticeable

estliberitas:
Make API link upper-cased so it's more simple to find it when browsing module page.


rmg:
Looks like this landed in b1b00ba641c8016c44cac21a187faea982bdd00c a few days ago :-)

PR #253, opened 2016-04-20
Update nyc to v6.4.0.

jeresig:
`nyc` v6.x added some useful new features, including the ability to include specific file extensions (such as `.jsx`). This updates `tap` to the latest version of `nyc` so that we can take advantage of them!

I'm using this modified version of `tap` in my application and it's working as expected.


isaacs:
Landed on e6a0f5672a8b9114fdee81aefaa0491e6e5b383c.  Thanks!

PR #256, opened 2016-05-09
Added types definition for TypeScript (picking up #238)

dlmanning:
I've picked up on @brennanMKE's work from #238 and attempted to address the points of @isaacs's feedback.
- No internal implementation details are exported
- All assertions and synonyms mentioned in the public documentation are included
- parameter types have been tightened down where possible
- Exported description now closely mirrors the actual exported object

Additionally, I've added a `typings` field to the package.json, which should make tap work "out of the box" with versions of Typescript >= 1.6. It's in a separate commit in case that weirds anyone out. (See: http://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html for reference)

Have attempted to be thorough and have referenced both the documentation and the source. Seems to work well from my own usage. Let me know if you see any problems and I'm happy to bang on it a little more. A few things I know of that are lacking / loose-endy:
1. There are some generated iterations of assertion synonyms not mentioned in the API docs that I didn't include in the type definitions. It wouldn't be a big deal to add them if you'd like
2. I've included only a super-minimal description of the promises returned from the `test`, `spawn` and `stdin` methods. What's actually returned is a Bluebird promise. Wasn't sure the right way to proceed with this.
3. I've made no attempt to document the inherited properties of readable-streams.

Cheers!

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "author": "Isaac Z. Schlueter <i@izs.me> (http://blog.izs.me)",
   "description": "A Test-Anything-Protocol library",
   "homepage": "http://node-tap.org/",
+  "typings": "./tap.d.ts",
   "bin": {
     "tap": "bin/run.js"
   },
--- a/tap.d.ts
+++ b/tap.d.ts
@@ -1,0 +1,39 @@
+// tap.d.ts
+
+declare module Test {
+
+  class Deferred {
+    resolve();
+    reject();
+  }
+
+  class test {
+    constructor(options?: any);
+    tearDown();
+    setTimeout(n: number);
+    endAll();
+    threw(er: Error, extra: Error, proxy: any);
+    pragma(set: any);
+    plan(n: number, comment?: string);
+    test(name: string, extra: Error, cb: () => void, deferred: Deferred);
+    loop(self: test, arr: test[], cb: () => void, i: number);
+    next(er: Error);
+    current();
+    stdin(name: string, extra: any, deferred: Deferred);
+    spawn(cmd: string, args: string, options: any, name: string, extra: any, deferred: Deferred);
+    done(implicit: boolean);
+    autoend();
+    printResult(ok: boolean, message: string, extra: any);
+    writeDiags(extra: any);
+    passing();
+    pass(message: string, extra: any);
+    fail(message: string, extra: any);
+    addAssert(name: string, length: number, fn: () => void);
+    comment();
+    push(chunk: any, encoding?: string);
+    bailout(message: string);
+    beforeEach(fn: () => void);
+    afterEach(fn: () => void);
+  }
+  
+}
--- a/tap.d.ts
+++ b/tap.d.ts
@@ -22,16 +22,13 @@ declare class Test {
   stdin(name: string, extra?: Options.Bag): Promise
   spawn(cmd: string, args: string, options?: Options.Bag, name?: string, extra?: Options.Spawn): Promise
   done()
-  autoend()
-  printResult(ok: boolean, message: string, extra: Options.Assert)
-  writeDiags(extra: Options.Bag)
   passing()
   pass(message?: string, extra?: Options.Assert)
   fail(message?: string, extra?: Options.Assert)
   addAssert(name: string, length: number, fn: (...args) => boolean)
   comment(message: string, ...args)
   bailout(message?: string)
-  beforeEach(fn: (cb: () => any) => void)
+  beforeEach(fn: (cb: () => any) => Promise | void)
   afterEach(fn: (cb: () => any) => Promise | void)
 
   // Assertions
--- a/tap.d.ts
+++ b/tap.d.ts
@@ -10,26 +10,26 @@ declare interface Tap extends Test {
 
 declare class Test {
   constructor (options?: Options.Test)
-  tearDown(fn: (any) => any)
-  setTimeout(n: number)
-  endAll()
-  threw(er: Error, extra?: Error, proxy?: Test)
-  pragma(set: Options.Pragma)
-  plan(n: number, comment?: string)
+  tearDown(fn:(a:any) => any):void
+  setTimeout(n: number):void
+  endAll():void
+  threw(er: Error, extra?: Error, proxy?: Test):void
+  pragma(set: Options.Pragma):void
+  plan(n: number, comment?: string):void
   test(name: string, extra?: Options.Test, cb?: (t: Test) => Promise | void): Promise
   test(name: string, cb?: (t: Test) => Promise | void): Promise
-  current()
+  current(): Test
   stdin(name: string, extra?: Options.Bag): Promise
   spawn(cmd: string, args: string, options?: Options.Bag, name?: string, extra?: Options.Spawn): Promise
-  done()
-  passing()
-  pass(message?: string, extra?: Options.Assert)
-  fail(message?: string, extra?: Options.Assert)
-  addAssert(name: string, length: number, fn: (...args) => boolean)
-  comment(message: string, ...args)
-  bailout(message?: string)
-  beforeEach(fn: (cb: () => any) => Promise | void)
-  afterEach(fn: (cb: () => any) => Promise | void)
+  done():void
+  passing():boolean
+  pass(message?: string, extra?: Options.Assert):boolean
+  fail(message?: string, extra?: Options.Assert):boolean
+  addAssert(name: string, length: number, fn: (...args:any[]) => boolean): boolean
+  comment(message: string, ...args:any[]):void
+  bailout(message?: string):void
+  beforeEach(fn: (cb: () => any) => Promise | void):void
+  afterEach(fn: (cb: () => any) => Promise | void):void
 
   // Assertions
   ok: Assertions.Basic
@@ -165,10 +165,10 @@ declare namespace Assertions {
     (obj: any, message?: string, extra?: Options.Assert) => boolean
 
   export type Throws =
-    (fn?: (any) => any, expectedError?: Error, message?: string, extra?: Options.Assert) => boolean
+    (fn?: (a:any) => any, expectedError?: Error, message?: string, extra?: Options.Assert) => boolean
 
   export type DoesNotThrow =
-    (fn?: (any) => any, message?: string, extra?: Options.Assert) => boolean
+    (fn?: (a:any) => any, message?: string, extra?: Options.Assert) => boolean
 
   export type Equal =
     (found: any, wanted: any, message?: string, extra?: Options.Assert) => boolean
@@ -180,7 +180,7 @@ declare namespace Assertions {
     (found: any, pattern: any, message?: string, extra?: Options.Assert) => boolean
 
   export type Type =
-    (found: any, type: string | ((any) => any), message?: string, extra?: Options.Assert) => boolean
+    (found: any, type: string | ((a:any) => any), message?: string, extra?: Options.Assert) => boolean
 }
 
 // Super minimal description of returned Promise (which are really Bluebird promises)
@@ -191,8 +191,8 @@ declare interface Promise {
 }
 
 declare interface Mocha {
-  it: (name?: string, fn?: (any) => any) => void
-  describe: (name?: string, fn?: (any) => any) => void
+  it: (name?: string, fn?: (a:any) => any) => void
+  describe: (name?: string, fn?: (a:any) => any) => void
   global: () => void
 }
 
@@ -203,3 +203,4 @@ declare interface TestConstructor {
 }
 
 export = tap;
+
--- a/tap.d.ts
+++ b/tap.d.ts
@@ -1,39 +1,208 @@
 // tap.d.ts
 
-declare module Test {
+declare var tap: Tap;
 
-  class Deferred {
-    resolve();
-    reject();
-  }
-
-  class test {
-    constructor(options?: any);
-    tearDown();
-    setTimeout(n: number);
-    endAll();
-    threw(er: Error, extra: Error, proxy: any);
-    pragma(set: any);
-    plan(n: number, comment?: string);
-    test(name: string, extra: Error, cb: () => void, deferred: Deferred);
-    loop(self: test, arr: test[], cb: () => void, i: number);
-    next(er: Error);
-    current();
-    stdin(name: string, extra: any, deferred: Deferred);
-    spawn(cmd: string, args: string, options: any, name: string, extra: any, deferred: Deferred);
-    done(implicit: boolean);
-    autoend();
-    printResult(ok: boolean, message: string, extra: any);
-    writeDiags(extra: any);
-    passing();
-    pass(message: string, extra: any);
-    fail(message: string, extra: any);
-    addAssert(name: string, length: number, fn: () => void);
-    comment();
-    push(chunk: any, encoding?: string);
-    bailout(message: string);
-    beforeEach(fn: () => void);
-    afterEach(fn: () => void);
-  }
-  
+declare interface Tap extends Test {
+  Test: TestConstructor
+  mocha: Mocha
+  mochaGlobals: () => void
 }
+
+declare class Test {
+  constructor (options?: Options.Test)
+  tearDown(fn: (any) => any)
+  setTimeout(n: number)
+  endAll()
+  threw(er: Error, extra?: Error, proxy?: Test)
+  pragma(set: Options.Pragma)
+  plan(n: number, comment?: string)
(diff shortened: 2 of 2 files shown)


isaacs:
This looks good, and is reasonably intelligible.

Can you add a test for it?  How would I verify that I haven't broken it (for example, by changing the code in the future to change an interface)?

dlmanning:
@isaacs : To test it I'll need to add typescript as a dev dependency. You cool with that?

isaacs:
Yeah, as a dev dep it's totally fine.

brennanMKE:
👍 Thanks for continuing with this update. I prefer `tap` over other test modules.

MelleB:
Compiling with `noImplicitAny` gives me the following errors. Might be a good idea to fix those before merge.

```
typings-custom/tap.d.ts(13,3): error TS7010: 'tearDown', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(13,17): error TS7006: Parameter 'any' implicitly has an 'any' type.
typings-custom/tap.d.ts(14,3): error TS7010: 'setTimeout', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(15,3): error TS7010: 'endAll', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(16,3): error TS7010: 'threw', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(17,3): error TS7010: 'pragma', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(18,3): error TS7010: 'plan', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(21,3): error TS7010: 'current', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(24,3): error TS7010: 'done', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(25,3): error TS7010: 'passing', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(26,3): error TS7010: 'pass', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(27,3): error TS7010: 'fail', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(28,3): error TS7010: 'addAssert', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(28,48): error TS7019: Rest parameter 'args' implicitly has an 'any[]' type.
typings-custom/tap.d.ts(29,3): error TS7010: 'comment', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(29,28): error TS7019: Rest parameter 'args' implicitly has an 'any[]' type.
typings-custom/tap.d.ts(30,3): error TS7010: 'bailout', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(31,3): error TS7010: 'beforeEach', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(32,3): error TS7010: 'afterEach', which lacks return-type annotation, implicitly has an 'any' return type.
typings-custom/tap.d.ts(168,12): error TS7006: Parameter 'any' implicitly has an 'any' type.
typings-custom/tap.d.ts(171,12): error TS7006: Parameter 'any' implicitly has an 'any' type.
typings-custom/tap.d.ts(183,35): error TS7006: Parameter 'any' implicitly has an 'any' type.
typings-custom/tap.d.ts(194,29): error TS7006: Parameter 'any' implicitly has an 'any' type.
typings-custom/tap.d.ts(195,35): error TS7006: Parameter 'any' implicitly has an 'any' type.
```

MelleB:
Created a pull request (for the pull request)
https://github.com/dlmanning/node-tap/pull/2

isaacs:
@dlmanning @MelleB This needs tests before I can land it.  Thanks.

dlmanning:
@isaacs @MelleB : I'll take care of it when I can find a bit of free time (hah), otherwise I definitely encourage anyone else who really wants this to take a swing.

Note that verifying the type definitions is probably at minimum an extremely verbose set of tests. Does anybody have an example of any other project that's implemented this sort of test?

benb:
Great stuff, I'm trying this right now. It seems to be missing support for `t.end()` though.

goodmind:
It seems that `t.jobs` is also don't working (neither if I add it to typings). It just throws

```ts
Cannot assign to 'jobs' because it is a constant or a read-only property. (2540)
```

philkunz:
What is the status of this? Types for tap are highly anticipated :D

philkunz:
@isaacs Any update on types for tap? :D

isaacs:
@philkunz If there are updates, they'll likely show up in this thread.

I'm waiting on a PR with tests that pass.

You might be able to get some assistance from the DefinitelyTyped folks.  Post a request here, maybe? https://github.com/DefinitelyTyped/DefinitelyTyped/labels/Definition%3ARequest

goodmind:
@isaacs but tests are passing
image
Or are you talking about tests for typings when some interface changes?

isaacs:
@goodmind Yeah, I need a test that fails without this file, and then passes with it, so that I don't accidentally break the typings in the future.

philkunz:
I see. That would require writing the tests themselves in typescript and transpiling them for type checking before the test. So if you break something the test's transpilation would fail or the test during runtime.

Another thought => Since TypeScript picks up speed and Angular, RxJS and soon Polymer will use it: How about using TypeScript for tap itself?  It makes sure typings are always up to speed (since autogenerated by typescript) And any valid JS is valid TypeScript, so there wouldn't be much refactoring required for making the switch.

goodmind:
@philkunz you don't need to rewrite tests in TypeScript. You just need one test in TypeScript for typings

philkunz:
@goodmind Well, you need to test every function that has typings in place? Since I suspect you don't test a lot twice, you should have all tests in TypeScript, don't you? Anyway that basically just is changing the tests extension  from .js to .ts ...

isaacs:
I'm not interested in rewriting node-tap or all of tap's tests in typescript.  (It's a lovely language, but I would rather keep tap in Plain Old JavaScript.  Call me a stick in the mud, it's fine.  I will be moving to node >=4 soon to use es classes and arrow functions, though.)

I just need a test file that loads tap, loads the typings, says "is this thing valid and are there obviously missing members/etc."  It doesn't need 100% coverage, it just needs to fail if the typings are actively broken.  (That probably will require generating a few test objects, of course.)  Otherwise, there's a very high chance that I or someone else will accidentally break the type definitions in the future, and I'd rather not have that on my conscience.  If it fails when I break it, I'll notice and fix it.

If someone else wants to take ownership for maintaining the types elsewhere (eg, in https://github.com/DefinitelyTyped/DefinitelyTyped) then I'm 100% ok with that.  But unless they follow this repo closely, there's a good chance we'll go out of sync over time, and that probably won't be great.

styfle:
Might I suggest creating a simple test such as the one [tape/index.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/tape/index.d.ts) has called [tape/tape-tests.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/tape/tape-tests.ts). I don't know how @isaacs refactors code but doing a replace-all (for example renaming an exported method) would replace both files and build. If only one file was changed, then it would not build and emit an error.

huan:
@isaacs I understand we must have a test for the new typing definition.

Could you accept the suggestion from @styfle?

It seems [tape/tape-tests.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/tape/tape-tests.ts) is a good way to make the typing definition clearly for maintaining(and sync). However, if you only change the interface of javascript, it will not fail. it will only be able to check the typing when you put the new interface to this test file.

robotlolita:
Starting with TypeScript 2.3 you can just point the compiler directly to JS files and have the compiler type check those. If you do so, you can just point the compiler directly to the existing test files. See https://github.com/Microsoft/TypeScript/wiki/Type-Checking-JavaScript-Files

The compiler invocation'd look like:

    ./node_modules/.bin/tsc --allowJs --checkJs --noImplicitAny --noEmit test/*.js

(the `--noEmit` means it'll just type check the files rather than compiling them and generating some output somewhere).

You'd also have to add some type annotations (in the form of comments) to the test files though, but if the type definitions are correct that'd mean just adding a type to relative requires. TypeScript can infer those require types iff there are specific type definitions for that particular module, but type definitions are usually only added to the entrypoint of a package.

Type inference should take care of the rest (unless meta-programming techniques are used through the tests), and removing implicit `any` types will make it fail if inference can't work (which usually means the types are wrong). e.g.:

```js
/// <reference path="../tap.d.ts" />
/** @type { Test } */
var test = require('../lib/root.js').test
```

The other alternative is just adding a single test file that uses all of the API. With that tests would break if existing API changes, but not if new things are added to the API.

isaacs:
I will literally land the first PR that has typings and a test for them.  I've been saying this for a year now.  Send the PR and I'll land it.  I don't know how to make it any more clear that a patch is welcome here.

You there!  Be the tap-using typescript-writing hero this community needs!

reverofevil:
These typings are wrong. `Test`  is not a class, it's a field of some type that resembles a class. Code with `test: Test` doesn't compile.

isaacs:
`Test` is as much of a class as JavaScript-circa-node-0.x has.  It's not using the `class` keyword, because that didn't exist, but it is a `function` with a prototype, and can be invoked using `new Test`.  It even inherits from a Base class :)

I do intend to drop support for node 0.x in the next major release, and refactor to use ES6 classes.  Will that make a difference for the type definitions?

isaacs:
(Note that `t.test` (lowercase) is _not_ a class, it's a function that queues up a test run that may or may not involve the creation of a `Test` instance.)

reverofevil:
@isaacs No, that wouldn't make any difference. It's only a problem with typings. As @NaridaL suggested on TS gitter, there must be an exported type, not a class. The [minimal example](http://stackoverflow.com/q/44027769/1872046) with required changes is here.

@dlmanning

kenhowardpdx:
The type definitions work just fine in the current form but it is missing some stuff addressed here (https://github.com/dlmanning/node-tap/pull/1). What @isaacs is looking for with this PR is a test file like this:

`tap-tests.ts`
```
import * as tap from 'tap';
declare var fs; // not testing
declare var __dirname; // not testing

let mam = function (x: number) {
  if (x % 2 === 0) {
    return 'even'
  } else if (x % 2 === 1) {
    return 'odd'
  } else if (x > 100) {
    return 'big'
  } else if (x < 0) {
    return 'negative'
  }
} 

tap.pass('this is fine');
tap.pass('this is fine', { skip: true, todo: 'true' });

tap.equal(mam(1), 'odd');
tap.equal(mam(2), 'even');
tap.equal(mam(200), 'big');
tap.equal(mam(-10), 'negative');

tap.test('some async stuff', function (childTest) {
  fs.readdir(__dirname, function (er, files) {
    if (er) {
      throw er // tap will handle this
    }
    childTest.match(files.join(','), /\basync\.js\b/)
    childTest.end()
  });
});

var t = new tap.Test();

t.equal(mam(1), 'odd');
```

I'm not familiar with this lib, but essentially all you need to do is add this file and go through the documentation and exercise each variation. If `tsc` compiles the test file without throwing an error then the tests have passed. The lib is very well documented so it shouldn't be hard... but it will take time.

reverofevil:
@kenhowardpdx Type definitions do not work just fine.

isaacs:
As discussed in #387, the requirement for me to accept typedefs in this module is that there is a test which fails whenever the typedef and the lib are out of sync.

That means that the tests cannot pass if the typedef:

1. defines a method that doesn't exist
2. defines return/arg/property types incorrectly
3. fails to define a public method/property
4. does not require rewriting the module or all tests in typescript

I am sure that it seems like I'm being overly picky about this, but I'd rather just not support typescript than be responsible for something that works incorrectly and is untested.  I'll break it, and I won't notice, and then I'll feel bad about it.  Better to just handle this with a typedef in a @DefinitelyTyped repo, so that users understand that they're not getting an authoritative reference from the module maintainer.

dlmanning:
For anyone else reading this thread and thinking of trying to meet @isaacs’s criteria. It probably isn’t impossible (though I’m not 100% sure some theoretical edge case won’t mean it is), but the level of type reflection and inspection he’s describing would probably be more complex than actually rewriting node-tap in TypeScript. Static types are designed to be maintained in concert with the rest of the source. @isaacs doesn’t want to do that. Fair enough, this is his show. Anyone considering putting more work into the type definitions should probably do it elsewhere.

Though if anyone out there is determined enough to actually make a system for fulfilling his requirements, I hope they’ll publish that work as it’s own module. :D

isaacs:
@dlmanning I just don't want to have someone working in VSCode (or some other TypeScript-aware editor) being told by the system that they can call `t.current()` to get a `Test` instance, but actually, that throws `TypeError: t.current is not a function`.  Incorrect typedefs are worse than none at all, and I'm not interested in shipping code that I _know_ I'll inadvertently break.

If that's not possible with the current state of the art in strongly-typed JavaScript, then ok.

Rewriting node-tap in typescript probably _would_ achieve the goal, because the type definitions would be inferred from the source itself, rather than a standalone artifact that can get out of sync.

> Though if anyone out there is determined enough to actually make a system for fulfilling his requirements, I hope they’ll publish that work as it’s own module. :D

Me too!  I'd gladly add type definitions to all of my modules if it were possible to test that they're in sync.

dlmanning:
@isaacs yeah, I’m agreeing with you. If the maintainer isn’t into keeping up the type definitions or actually writing the project in TypeScript, it’s probably a bad idea to include them. The sort of tests you’ve described would require hacking into the TypeScript compiler just to get started. It’d be an interesting enough endeavor to definitely deserve its own project.

reverofevil:
@isaacs As anyone can still download these typings and fix them appropriately to the current version, it's fine. Stuff with wrong intellisense in VSCode is plain creepy, and it made me think if I ever called non-existing methods in production code. (The point wasn't this clear before.)

Please, consider making an article on this topic to gather more attention.

NoNameProvided:
@isaacs there is now a way to check them via [Microsoft/dtslint](https://github.com/Microsoft/dtslint)

Eoksni:
There is a library for runtime checks of type validity https://github.com/gcanti/io-ts. I'm excited to try it out and writing tests for tap types looks like a perfect candidate for that. If @isaacs is still willing to merge types when tests are landed then I'll try to make it work. For sure it can validate the *shapes* of public interfaces, but validating function argument and return types might be more challenging.

isaacs:
Just to reiterate: I'm fine with this as long as it's maintainable.  First and foremost, it must have a test that fails whenever the type definitions are incorrect, missing, or otherwise inadequate, and they must not be too hard to keep up to date.  I'm probably not going to look at them at all except when the test fails.  If that's too hard to do, just send it to DefinitelyTyped.

phaux:
@isaacs There's now a way to generate TypeScript definitions from JS 🎉 Just add some JSDoc comments to the lib and that's all.

https://github.com/microsoft/TypeScript/pull/32372

Definitions will be always in sync. Does it address your concern?

iffy:
I'll point this out again in case it's still useful.  It meets isaacs criteria without changes to the code: https://github.com/tapjs/node-tap/pull/381  It just needs to be rebased onto master.

I'm no longer using this, so I'll be unsubscribing, but if anyone wants to pick it up, I'm pretty sure it would solve this issue.

isaacs:
I'm fine with JSDoc comments.  Tests should fail if they're out of sync, though.

weswigham:
> I'm fine with JSDoc comments. Tests should fail if they're out of sync, though.

Your test, in such a scenario (once the js declaration file support is in), would then simply be running `tsc --declaration --emitDeclarationOnly --declarationDir ./types --allowJs --checkJs --target es6 --module commonjs ./lib/tap.js`, which would fully typecheck your codebase using the doc comments as the source of type information (and also generate the declaration files).

I actually went ahead and tried it out [here](https://github.com/weswigham/node-tap/tree/add-generate-definitions) with an early version of the generator to see how we do - the types we make are OK (there's a method-substitutes-for-property bug in the output declarations I need to look into, probably), but pretty much every parameter is untyped (they need jsdoc comments specifying their types), and there's a small handful of patterns the compiler needs assistance with via some extra `@type` jsdoc annotations (or are maybe legitimate errors, hard to tell, as I'm unfamiliar with the project).

isaacs:
Sounds good.  Patch welcome :)

PR #257, opened 2016-05-10
add node v6 to travis conf

a0viedo:


huan:
will tap run ok under nodejs 6.0 or above right now?

isaacs:
Node 6.0 changed stdio behavior in a way that makes tap (and many other cli apps) not function as expected.  It's not yet stable.  When it is, I'll add it to the test config.  For now, it's not worth the effort to try to make it work.

PR #263, opened 2016-05-31
Add CLI flag to avoid opening a browser window

dotproto:
I've added a new command line flag to prevent tap from opening a new browser window after generating an HTML report.

I wanted to use nodemon with my project's test-coverage npm script, but the auto-open behavior ended up spawning a lot of unused browser tabs. Hopefully this will address that workflow issue should anyone else run into a similar hitch.

I couldn't find anywhere in the existing tests for command line options. Please let me know if I missed something obvious or if I should take a crack at creating a new test.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -141,6 +141,7 @@ function parseArgs (args) {
   var defaultCoverage = options.pipeToService
 
   options.coverageReport = null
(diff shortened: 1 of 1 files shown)


isaacs:
Landedon master, will be published in 5.8.0 soon.

PR #268, opened 2016-06-07
patch: add hasOwnProperty checks into for...in loops to prevent iteration over unexpected members

schwarzkopfb:
patch: add hasOwnProperty checks into for...in loops to prevent iteration over unexpected members
docs: fix broken link in CONTRIBUTING.md


schwarzkopfb:
Prototypical extension of built-in classes (e.g. Object) is a known antipattern, however there are still people out there, who use this option. As a developer creating generic modules, I think testing a library against these extensions is a real use case.

isaacs:
So, there are a few things that are blocking landing this as-is.
1. The JS style should be 2-space indents everywhere.
2. The added test doesn't fail without the code changes, so I'm not exactly sure what it is testing?
3. In general, adding `for (x in y) { if (y.hasOwnProperty(x)) { ...` is not ever the best approach.  Either the prototype chain is relevant, or it is not.  If it is, then the `hasOwnProperty` check is incorrect.  If it is not, then `Object.keys(y).forEach(function (x) { ... })` is a better pattern.  Furthermore, if the prototype chain is not trustworthy for some reason, or if the object is a generic bag o' data which can contain any key, then it's better to do `var hasOwnProperty = Object.prototype.hasOwnProperty` at the top of the file, and then test with `hasOwnProperty.call(obj, key)` (since the `hasOwnProperty` key may be overwritten, or not a function, or some other function, etc.)

If you feel like rebasing onto the latest master, and fixing these things, I'd be happy to merge this.  Please let me know if I can be of any assistance.

Thanks!

schwarzkopfb:
Thank you for your response!

I realised that my original issue has been fixed meanwhile by the replacement of `for...in` here: https://github.com/tapjs/node-tap/blob/master/lib/assert.js#L189-L191

However, tap still breaks if I `require()` it _after_ `Object.prototype` has been extended, because of this loop: https://github.com/tapjs/node-tap/blob/master/lib/synonyms.js#L67

My latest commit on this fork fixes that. I also replaced the `for...in` loops everywhere in the code base with `Object.keys(y).forEach(function (x) { ... })` to prevent any further issues like this. Sorry for the indentation mistake in my test file (that you mentioned in point 1.), it is also fixed now.

I had difficulties and conflicts during the rebasing process, so please double-check my changes, but I think everything is ok.

lib/mocha.js
```diff
@@ -7,9 +7,11 @@ var stack = require('./stack.js')
 exports.it = it
 exports.describe = describe
 exports.global = function () {
-  for (var g in exports) {
-    global[g] = exports[g]
-  }
+  Object.keys(exports).forEach(function (g) {
+    if (exports.hasOwnProperty(g)) {
```
isaacs: Unnecessary check.  `Object.keys` only loops over own enumerable properties.

```
> var parent = { parent: true }
undefined
> var child = Object.create(parent, { child: { value: true, enumerable: true }, other: { value: 1 }})
undefined
> Object.keys(child)
[ 'child' ]
```

lib/test.js
```diff
@@ -108,12 +108,12 @@ function Test (options) {
 
   // Bind all methods.
   var bound = {}
-  for (var m in this) {
+  Object.keys(this).forEach(function (m) {
```
isaacs: Semantic change.  The intent here is to bind inherited as well as local functions.  (Generally, all or most methods are properties of the prototype, not local properties.)

lib/test.js
```diff
@@ -347,9 +347,11 @@ Test.prototype.pragma = function (set) {
     return
   }
 
-  for (var i in set) {
-    this.push('pragma ' + (set[i] ? '+' : '-') + i + '\n')
-  }
+  Object.keys(set).forEach(function (i) {
+    if (set.hasOwnProperty(i)) {
```
isaacs: Unnecessary.

lib/test.js
```diff
@@ -347,9 +347,11 @@ Test.prototype.pragma = function (set) {
...
+      this.push('pragma ' + (set[i] ? '+' : '-') + i + '\n')
+    }
+  })
```
isaacs: `this` will be different in the forEach method.  This line should be `}, this)` to explicitly bind the forEach iterator.

test/test/has-own-property.js
```diff
@@ -0,0 +1,14 @@
...
+var tap = require('../..')
+
+function error() {
+  throw new Error('fake')
+}
+
+tap.test('obj.hasOwnProperty(key) checks', function (test) {
+  test.throws(error, 'this fails without internal checks')
+  test.end()
+})
```
isaacs: There's no need for an output check in test/test/.  A top-level test at `test/has-own-property.js` would be sufficient, and would only have to do this:

``` javascript
Object.prototype.foo = 42
require('../').pass('modifying Object.prototype did not break things')
```

schwarzkopfb:
Unsurprisingly, your observations were perfectly valid, so I made the proposed changes. Is there anything else that's blocking this request to be merged?

PR #302, opened 2016-09-03
Fixing endless loop for coverage if -- is used when run in node_modules

martinheidegger:
My tests were run in a constellation where [`nyc` didn't set the `global.__coverage__`](https://github.com/tapjs/node-tap/blob/e49616d1e8c7f726e1c6834bbec81c19509f72fa/bin/run.js#L240) property and I used `--cov -- mytest.js` as command line options.

In this constellation node-tap ran into an **endless loop** because all arguments after `--` are treated as [files](https://github.com/tapjs/node-tap/blob/e49616d1e8c7f726e1c6834bbec81c19509f72fa/bin/run.js#L366), including `--__coverage__`.

This PR fixes that problem by testing the arguments before any other parsing for `__coverage__`.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -173,6 +173,17 @@ function parseArgs (args, defaults) {
   var defaultCoverage = options.pipeToService
   var dumpConfig = false
 
+  if (args.indexOf('--__coverage__') !== -1) {
+    // NYC will not wrap a module in node_modules.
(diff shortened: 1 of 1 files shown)


isaacs:
I think this is better handled by using an environment variable.  I have a fix that I'll land in the next patch release.

Issue #305, opened 2016-09-06
Buffered subtests


isaacs:
``` js
t.test('test name', { buffer: true }, function (t) {
  t.plan(2)
  t.pass('this is fine')
  t.equal(1, 1)
})
```

This buffers the output until the test finishes, and produces output like:

``` tap
ok 1 - test name {
    1..2
    ok 1 - this is fine
    ok 2 - should be equal
}
```

isaacs:
Landed in v9

PR #309, opened 2016-09-06
Buffered tests

isaacs:
Just a wip for #305.  Do not land yet.
- [ ] handle bailouts nicer (right now it's kind of noisy, better to just close the `}` and then bail out at top level, like with streaming subtests)
- [ ] code is a bit kludgey, clean that up
- [x] unfinished tests need to get closed properly (right now it just exits uselessly)
- [ ] support stdin as a buffered subtest
- [x] verify buffered fixture output in test/test.js
- [x] mark empty buffered tests (ie, with `1..0`) as `TODO`, maybe?
- [ ] bailOnFail should not bail out on tests with a `{` until the next thing at that indentation level, so that the bailout happens at the error, not before the actual error is printed.


isaacs:
These tests produce no output, and should at least produce something.

```
$ find test/test/*-buffer* -size 1c
test/test/empty-bail-buffer.tap
test/test/empty-buffer.tap
test/test/plan-too-many-bail-buffer.tap
test/test/plan-too-many-buffer.tap
test/test/throw-bail-buffer.tap
test/test/throw-buffer.tap
test/test/unfinished-bail-buffer.tap
test/test/unfinished-buffer.tap
```

Issue #342, opened 2017-02-01
Add -J to tap to auto calculate the cores


gdams:
It would be really great if we could add a feature to tap where instead of using `tap -j<n>` we could use `tap -J` and it would fetch the number of cores from `os.cpus().length`. The issue with having to pass in the number of cores is that it makes it very hard to implement parallel tap into a package.json

isaacs:
This is a good idea.  Patch welcome, should be entirely doable in `bin/run.js`.

PR #344, opened 2017-02-01
Add -J to automatically calculate cores

gdams:
This PR allows you to pass in `-J` which will automatically calculate
the number of cores using `os.cpus().length`

Fixes: https://github.com/tapjs/node-tap/issues/342

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -13,6 +13,7 @@ var osHomedir = require('os-homedir')
 var yaml = require('js-yaml')
 var path = require('path')
 var exists = require('fs-exists-cached').sync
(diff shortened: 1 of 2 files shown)


gdams:
CC @gibfahn

Fishrock123:
This would be great. :D

isaacs:
I think `--Jobs` is too close to `--jobs` and doesn't describe what it does.  I'm replacing it with `--jobs-auto`, keeping `-J` as the shorthand.

Landed on f3f0a6b2b9e506a01bd90b716fb9f5c69c7c942c.

PR #371, opened 2017-06-02
Add NaN support for strictSame()

nechaido:
`deeper` support seems to be discontinued, last commit is of Aug 31, 2015.
Also `lodash.isequal` handles `NaN` properly but `deeper` doesn't, which is the main reason of this PR.


isaacs:
Fixed on 10.5.1.

PR #374, opened 2017-06-10
WIP – TAP_ONLY, --only, {only: true} and "only" warning

### LINKED ISSUE #187: Exclusive test, .only method

gr2m:
# THIS IS WORK IN PROGRESS

writes warning to stderr if {only: true} but tap’s only option is not true

closes #187 

### Progress

- [x] make existing tests pass (rebase on master https://github.com/tapjs/node-tap/pull/374#issuecomment-309129473)
- [x] create test in `test/test/...` https://github.com/tapjs/node-tap/pull/374#issuecomment-308904189
- [x] create runner test https://github.com/tapjs/node-tap/pull/374#issuecomment-308904189
- [x] add --only to the set of configs that can be loaded from a .taprc file https://github.com/tapjs/node-tap/pull/374#issuecomment-308904189
- [x] update `usage.txt`

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -110,7 +110,11 @@ Test.prototype.spawn = function spawn (cmd, args, options, name) {
 }
 
 Test.prototype.sub = function (Class, extra, caller) {
-  if (extra && (extra.todo || extra.skip)) {
+  if (extra.only && !this.only) {
+    process.stderr.write('# Warning: "' + extra.name + '" has {todo: true} set but all tests are run\n')
+  }
+
+  if (extra && (extra.todo || extra.skip || (this.only && !extra.only))) {
     this.pass(extra.name, extra)
     return Promise.resolve(this)
   }
--- a/lib/test.js
+++ b/lib/test.js
@@ -111,7 +111,7 @@ Test.prototype.spawn = function spawn (cmd, args, options, name) {
 
 Test.prototype.sub = function (Class, extra, caller) {
   if (extra.only && !this.only) {
-    process.stderr.write('# Warning: "' + extra.name + '" has {todo: true} set but all tests are run\n')
+    process.stderr.write('# Warning: "' + extra.name + '" has {only: true} set but all tests are run\n')
   }
 
   if (extra && (extra.todo || extra.skip || (this.only && !extra.only))) {
--- a/bin/run.js
+++ b/bin/run.js
@@ -154,6 +154,7 @@ function parseArgs (args, defaults) {
     C: 'no-color',
     T: 'no-timeout',
     J: 'jobs-auto',
+    O: 'only',
     h: 'help',
     '?': 'help',
     v: 'version'
--- a/bin/run.js
+++ b/bin/run.js
@@ -375,6 +375,10 @@ function parseArgs (args, defaults) {
         options.bail = false
         continue
 
+      case '--only':
+        options.only = true
+        continue
+
       case '--':
         options.files = options.files.concat(args.slice(i + 1))
         i = args.length
@@ -576,6 +580,9 @@ function setupTapEnv (options) {
 
   if (options.bail)
     process.env.TAP_BAIL = '1'
+
+  if (options.only)
+    process.env.TAP_ONLY = '1'
 }
 
 function globFiles (files) {
--- a/bin/usage.txt
+++ b/bin/usage.txt
@@ -37,6 +37,8 @@ Options:
 
   -B --no-bail                Do not bail out on first failure (Default)
 
+  -O --only                   Only run tests with {only: true} option
+
   -R<type> --reporter=<type>  Use the specified reporter.  Defaults to
                               'classic' when colors are in use, or 'tap'
                               when colors are disabled.
--- a/lib/base.js
+++ b/lib/base.js
@@ -42,6 +42,7 @@ function Base (options) {
   this.jobs = +ownOrEnv(options, 'jobs', 'TAP_JOBS') || 0
   this.skip = ownOr(options, 'skip', false)
   this.todo = ownOr(options, 'todo', false)
+  this.only = ownOr(options, 'only', false)
   this.setupParser(options)
   this.finished = false
   this.output = ''
@@ -212,6 +213,7 @@ Base.prototype.inspect = function () {
     output: this.output,
     skip: this.skip,
     todo: this.todo,
+    only: this.only,
     results: this.results,
     options: [
       'autoend',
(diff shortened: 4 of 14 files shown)


gr2m:
ping @isaacs I need your input here to know if this goes in the right direction before I can continue.
If you are too busy right now let me know when / how to ping again, happy to do so

isaacs:
Reading through this now, sorry for the delay :)

lib/test.js
```diff
@@ -110,7 +110,11 @@ Test.prototype.spawn = function spawn (cmd, args, options, name) {
 }
 
 Test.prototype.sub = function (Class, extra, caller) {
-  if (extra && (extra.todo || extra.skip)) {
+  if (extra.only && !this.only) {
+    process.stderr.write('# Warning: "' + extra.name + '" has {todo: true} set but all tests are run\n')
```
isaacs: Do you mean `{only: true}`?

isaacs:
This strikes me as the correct approach, overall.  (Apart from what I'm pretty sure is a typo in the warning message?)

Needs a test.  You could add a js file to `test/test/...` and then `npm run regen-fixtures` to verify output.  Probably there should be a runner test as well.

It looks like we'll probably also want to add `--only` to the set of configs that can be loaded from a `.taprc` file?

isaacs:
Not sure why the tests seem to be hanging on node v6.  I'll try to investigate later.

isaacs:
@gr2m If you rebase onto master, it should make the test failures go away.  `--debug` is super flaky on node 6, and removed in node 8, so I deleted the test, since it hangs about half the time.

gr2m:
I’ve rebased on top of master. The tests pass locally with node v6.9.5 (npm v3.10.10) and on travis with all versions now. Not sure why they fail on appveyor: https://ci.appveyor.com/project/isaacs/node-tap/build/405/job/4cfnem7oasdhwvy0

it looks like it’s not related to this PR though, as previous builds failed, too

> It looks like we'll probably also want to add --only to the set of configs that can be loaded from a .taprc file?

That seems to work already? Or am I missing something?

```bash
$ ./node_modules/.bin/tap test/only-test.js 

test/only-test.js # Warning: "only" has {only: true} set but all tests are run
test/only-test.js ..................................... 2/2
total ................................................. 2/2

  2 passing (294.111ms)

  ok

$ cat .taprc 
only: true

$ TAP_RCFILE=`pwd`/.taprc ./node_modules/.bin/tap test/only-test.js 

total ................................................. 1/1

  1 passing (7.752ms)

  ok
```

gr2m:
> Probably there should be a runner test as well 

Is this what you had in mind? 2febf88

gr2m:
This doesn’t look right yet. 

for a test file looking like this:

```js
var test = require('../..').test

test('root test without {only: true}', function (t) {
  t.pass('test will not run without --only')
  t.end()
})

test('root test with {only: true}', {only: true}, function (t) {
  t.pass('test will log an error when run without --only')
  t.end()
})

test('group', function (group) {
  group.test('subtest without {only: true}', function (t) {
    t.pass('subtest will not run without --only')
    t.end()
  })

  group.test('subtest with {only: true}', {only: true}, function (t) {
    t.pass('test will log an error when run without --only')
    t.end()
  })

  group.end()
})
```

I get this result when run without the `only` flag:

```bash
$ bin/run.js --reporter tap test/fixtures/only.js 
TAP version 13
# Subtest: test/fixtures/only.js
    # Subtest: root test without {only: true}
        ok 1 - test will not run without --only
        1..1
# Warning: "root test with {only: true}" has {only: true} set but all tests are run
    ok 1 - root test without {only: true} # time=5.29ms
    
    # Subtest: root test with {only: true}
        ok 1 - test will log an error when run without --only
        1..1
    ok 2 - root test with {only: true} # ONLY
    
    # Subtest: group
        # Subtest: subtest without {only: true}
            ok 1 - subtest will not run without --only
            1..1
        ok 1 - subtest without {only: true} # time=0.849ms
        
# Warning: "subtest with {only: true}" has {only: true} set but all tests are run
        # Subtest: subtest with {only: true}
            ok 1 - test will log an error when run without --only
            1..1
        ok 2 - subtest with {only: true} # ONLY
        
        1..2
    ok 3 - group # time=10.348ms
    
    1..3
    # time=32.296ms
ok 1 - test/fixtures/only.js # time=246.795ms

1..1
# time=260.614ms
```

And this result when run with the `only` flag

```
$ bin/run.js --only --reporter tap test/fixtures/only.js 
TAP version 13
ok 1 - test/fixtures/only.js
1..1
# time=8.261ms
```

But I would expect this output

```bash
$ bin/run.js --only --reporter tap test/fixtures/only.js 
TAP version 13
# Subtest: test/fixtures/only.js
    # Subtest: root test with {only: true}
        ok 1 - test will log an error when run without --only
        1..1
    ok 1 - root test with {only: true}
    
    # Subtest: group
        # Subtest: subtest with {only: true}
            ok 1 - test will log an error when run without --only
            1..1
        ok 1 - subtest with {only: true}
        
        1..1
    ok 2 - group # time=8.193ms
    
    1..2
    # time=30.159ms
ok 1 - test/fixtures/only.js # time=261.422ms

1..1
```

My expected output would be the same as the first, but it would leave out all tests without the `{only: true}` entirely.

Does that sound right to you @isaacs?

isaacs:
Yeah, I'd expect that the `--only` flag will set `tap.only` at the root, and then exclude any subtests that don't have `{ only: true }`, but still include sub-subtests of included tests.

Maybe you can do something like `this.only = parent ? parent.only : false` in the Test constructor?

Also, it's a little confusing to use the same flag to mean "only run this thing" and "this thing only runs child things with the 'only run this thing' flag".  Maybe consider `this.runOnly` instead of `this.only`?

isaacs:
@gr2m I'll have some time to review this more carefully this weekend, and I might be able to just fix the issues you're running into.

I think it's almost there, and I wanna get this and `--grep` landed before switching gears to work on v11.  Thanks for digging into it :)

isaacs:
@gr2m Here's what I'm talking about: https://github.com/tapjs/node-tap/commit/8a47fa238c1d70fc78cb19b2cdfe05382975d040

I haven't updated the tests, so it's probably failing, but it seems like this generates the intended result when running the fixture.

gr2m:
shipped in c41742440869f06ee1eac39d9fc80a3fc3af3f53. Thanks :)

PR #381, opened 2017-06-27
Add TypeScript definition

iffy:
This is picking up on Issue #256.  (Aside: Seems weird to make a different PR instead of tracking the conversation and changes in one place.  GitHub seems to not allow for otherwise, though)

I've added a "test" (i.e. compile a TypeScript test file).  But... I *think* that the TypeScript test file needs to exercise every function (and function variation) in order to be useful.  And my problem is two-fold:

1. I don't know how to use Tap yet (I'm just picking it up), so I don't know how to exercise every part of it.
2. I don't know how to exercise the test-failing parts (e.g. `tap.fail("this failed")`) without also making the tests fail.

Any ideas?

Issue #237, opened 2016-02-15
Creating TypeScript types definition


brennanMKE:
http://definitelytyped.org/guides/contributing.html

There's currently no definition for the `tap` module. I like using `tap` so I will start creating the initial definition for the functions that I am using. I can submit it as a PR and then follow the guidelines for contributing to DefinitelyTyped.

brennanMKE:
This may cover most of it.

``` typescript
// tap.d.ts

declare module Test {

  class Deferred {
    resolve();
    reject();
  }

  class test {
    constructor(options?: any);
    tearDown();
    setTimeout(n: number);
    endAll();
    threw(er: Error, extra: Error, proxy: any);
    pragma(set: any);
    plan(n: number, comment?: string);
    test(name: string, extra: Error, cb: () => void, deferred: Deferred);
    loop(self: test, arr: test[], cb: () => void, i: number);
    next(er: Error);
    current();
    stdin(name: string, extra: any, deferred: Deferred);
    spawn(cmd: string, args: string, options: any, name: string, extra: any, deferred: Deferred);
    done(implicit: boolean);
    autoend();
    printResult(ok: boolean, message: string, extra: any);
    writeDiags(extra: any);
    passing();
    pass(message: string, extra: any);
    fail(message: string, extra: any);
    addAssert(name: string, length: number, fn: () => void);
    comment();
    push(chunk: any, encoding?: string);
    bailout(message: string);
    beforeEach(fn: () => void);
    afterEach(fn: () => void);
  }

}
```

brennanMKE:
See PR

https://github.com/tapjs/node-tap/pull/238

isaacs:
I'm in favor of this!  Comment left on PR.  Thanks for taking this on :)

brennanMKE:
I haven't had time to make any updates. We are officially using Mocha but I have used `tap` in the past because it just works. I will try to come back to it for my own projects which I will be doing in TypeScript going forward.

jamestalmage:
> I can submit it as a PR and then follow the guidelines for contributing to DefinitelyTyped.

My understanding is that it is no longer necessary to submit to DefintelyTyped if the `.d.ts` file is deployed to `npm`. DefinitelyTyped still exists to accommodate libs that refuse to do that. [Reference](https://www.reddit.com/r/typescript/comments/3ur0xb/question_node_where_should_i_store_my_dts_files/)

brennanMKE:
[...]

PR #387, opened 2017-07-19
add(typescript): typing defination file with unit tests (#237)

huan:
This PR made from all the TypeScript resources from the Issues & Pull Requests before. Include: #237 #238 #256 #381 .

What this PR do is:
1. Added `index.d.ts` to the root directory, added `types` entry to `package.json`;
1. Added `test/typescript.ts` unit test to make sure the basic TypeScript system works right;
1. Added TypeScript linting tool to do **Static Type Checking**;
1. Integrate the TypeScript test & lint scripts to `package.json`, which can be run automatically by executing `npm test`;

I really want to use `import { test } from 'tap'` with the Type Checking / IDE IntelliSense, instead of `const { test } = require('tap')` and have to check the document from time to time.

@isaacs Please let me know if there need do any future works on it, Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Just kidding.
 Most frameworks spend a lot of their documentation telling you why
 they're the greatest.  I'm not going to do that.
 
-### <i lang="it">tutti i gusti, sono gusti</i>
+### <i lang="it">tutti i gusti sono gusti</i>
 
 Software testing is a software and user experience design challenge
 that balances on the intersection of many conflicting demands.
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,211 @@
+/**
+ * https://github.com/tapjs/node-tap/pull/256#issuecomment-246633973
+ * Credit: @MelleB @dlmanning @brennanMKE @iffy
+ */
+// tap.d.ts
+
+declare var tap: Tap;
+
+declare interface Tap extends Test {
+  Test: TestConstructor
+  mocha: Mocha
+  mochaGlobals: () => void
+}
+
+declare class Test {
+  constructor (options?: Options.Test)
+  tearDown(fn:(a:any) => any):void
+  setTimeout(n: number):void
+  endAll():void
+  end(implicit?: any): void
+  threw(er: Error, extra?: Error, proxy?: Test):void
+  pragma(set: Options.Pragma):void
+  plan(n: number, comment?: string):void
+  test(name: string, extra?: Options.Test, cb?: (t: Test) => Promise | void): Promise
+  test(name: string, cb?: (t: Test) => Promise | void): Promise
+  current(): Test
+  stdin(name: string, extra?: Options.Bag): Promise
+  spawn(cmd: string, args: string, options?: Options.Bag, name?: string, extra?: Options.Spawn): Promise
+  done():void
+  passing():boolean
+  pass(message?: string, extra?: Options.Assert):boolean
+  fail(message?: string, extra?: Options.Assert):boolean
+  addAssert(name: string, length: number, fn: (...args:any[]) => boolean): boolean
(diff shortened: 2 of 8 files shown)


huan:
TravisCI passed! :smiley:

ORESoftware:
nice work, I will be happy to use this thanks!

huan:
ping @isaacs

huan:
AppVeyor is failing because of they have a node runtime called `node.EXE` but we are checking for `node.exe`.

<https://ci.appveyor.com/project/isaacs/node-tap/build/446/job/46d8untfsgpj54yv#L5078>
```shell
              found: '    not ok 1 - node.EXE ./test/test/nesting.js # time=650.715ms'
              pattern: >-
                /^    not ok 1 \- .*(node|iojs)(.exe)? \.\/test\/test\/nesting\.js #
                time=[0-9.]+(ms)?$/
```

Will try to fix it.

huan:
Windows CI is passed after we ignore the case for `node.EXE` 

I believe this is the reason why the Appveyor CI looks not stable in the commit history. ;-]

huan:
ping @isaacs

huan:
Hi @isaacs, this PR have included all requirements that we discussed before, especially the unit tests.

I'd like to contribute more if we need additional works on it. If so, please feel free to let me know what you want me to do before this PR could be able to be merged.

I use tap in lots of my TypeScript projects, it's a pain for me that can not see any IDE IntelliSense and without any static typing check. 

Looking forward to hearing back from you, thanks!

kazimuth:
Ping, is there anything else that needs to be done for this to be merged?

isaacs:
Hi, sorry for the delay, thank you for your patience while v11 was being completed.

I've ported everything up to latest and squashed in my `ts` branch here: https://github.com/tapjs/node-tap/commit/ts

The only issue I still have with this is how am I to know that it's accurate and complete?  Even though many new methods are added on v11, it seems to be fine with them not being covered in the typescript definition.  Is there a way to make it fail on un-type-defined methods?

isaacs:
Ack, worse yet, the `Test.prototype.current()` method is removed, and the type definition checks are fine with it still being in the type definition.  That's exactly what I was afraid of.  I will break this thing, and not notice.  That's not ok.

Without automated testing, including automatic test coverage enforcement, I can't take responsibility for this.  I'm sorry.

I'm happy to merge this into master if that concern can be addressed somehow.

huan:
Hi @isaacs,

Glad to see you just released v11, and I understand that you are still worrying the mismatch between the Typing Definition and the JS code.

As my 2 cents, it will always be acceptable for me to have a buggy typing definition file instead of none. ;-p

However, if we want to be the perfect match, the ultimate solution would write the interface JS code to TS, and we use `tsc` to compile TS to JS and then we will have an absolutely matched type definition file.

For example, I just had a quick look at tap source code, it seems the API is inside the following two files:
1. https://github.com/tapjs/node-tap/blob/master/lib/test.js
1. https://github.com/tapjs/node-tap/blob/master/lib/tap.js

They are all written in ES6 class, which means it could be very easy to rewrite in TypeScript. If so, then we will be never worried about the mismatch anymore.

However, I do not know if this solution is acceptable for @isaacs. If you think it is acceptable, I'd like to send you a PR to demo this.

isaacs:
Yeah, I'm really not interested in migrating to typescript for this module right now.  Maybe the best approach is to take this over to https://github.com/definitelytyped or something.

isaacs:
According to @iffy in [this comment](https://github.com/tapjs/node-tap/pull/381#issuecomment-348976674), the branch used in #381 may meet the necessary criteria?  If anyone wants to take a look, I'm still open to landing this, provided there's checks against me breaking it later :)

ericmorand:
Any news on this?

PR #395, opened 2017-08-01
Fix Italian quote punctuation in readme

matteocontrini:
In the sentence `tutti i gusti, sono gusti`, the subject `tutti i gusti` is separated from the verb `sono` with a comma. That's an error in Italian. Punctuation cannot be placed between the subject and the verb.


isaacs:
Thanks!  I'd only ever heard it spoken.

ORESoftware:
Error, does not compile

PR #396, opened 2017-08-04
Add tapshot to 100 club

CodeFoodPixels:
Switched this module to use --100 the other day


isaacs:
This is a cool utility!  Someone pointed me at this feature in Jest recently, and I was keen to figure out a way to get it implemented in node-tap.  I'll definitely recommend tapshot as an option!

PR #403, opened 2017-09-08
Add --require option to the cli

viktor-ku:
Hey guys

I personally miss this feature very very much. I am having so much headache trying to figure out how to run awesome `node-tap` tests with e.g. babel

The reason to run src tests is to use flow in my case

Please tell me what I need to do to finish it ASAP and maybe include in v 11?

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -165,6 +165,7 @@ function parseArgs (args, defaults) {
   }
 
   var singleOpts = {
+    r: 'require',
     j: 'jobs',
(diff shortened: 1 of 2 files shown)


isaacs:
Please rebase onto current master, and add a test in `test/run.js` to exercise the code path and maintain 100% coverage.

isaacs:
This can be done today using `--node-arg=--require=babel`.  Punting for now, since there doesn't seem to be too much activity here.

PR #416, opened 2017-12-06
test: add emitsEvent() assertion

evanlucas:
This adds a new assertion that verifies that an EventEmitter emits an
event. 

I've written this quite a few times now, so I figured someone else may find it useful.
I do think having the ability to match against the argument(s) being emitted would be a nice to have as well, but wasn't sure on the naming. Anyways, I wanted to go ahead and see if you were interested in this before implementing matching.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -1348,6 +1348,56 @@ class Test extends Base {
         t.match(value, wanted, extra)
       }))
   }
+
+  emitsEvent (emitter, eventName, message, extra) {
+    this.currentAssert = Test.prototype.emitsEvent
+
+    if (message && typeof message === 'object')
+      extra = message, message = ''
+
+    if (!extra)
+      extra = {}
+
+    if (!message)
+      message = `expect emitter to emit event: "${eventName}"`
+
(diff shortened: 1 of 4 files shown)


evanlucas:
After looking more through the other assertions, it may make more sense to name this `emits` and then have `emitMatch` kinda like `resolves` and `resolveMatch`.

lib/test.js
```diff
@@ -1348,6 +1348,56 @@ class Test extends Base {
         t.match(value, wanted, extra)
       }))
   }
+
+  emitsEvent (emitter, event_name, message, extra) {
```
isaacs: camels, please, snake_case is too scary because snakes.

evanlucas: will do

isaacs:
@evanlucas Hey, I'm wondering if the approach in 38119eec131de4cda7451027cfc2d71d50f46f90 would satisfy your use case.  I'm not super keen on having a timer kicking around.  Better to just trigger it on test end.

evanlucas:
@isaacs it certainly does. Seems to be implemented more reliably too. Thanks for that!

PR #423, opened 2018-02-12
Update own-or-env to 1.0.1

jsumners:
This PR updates the `own-or-env` dependency to version 1.0.1. This fixes an issue running `tap` when projects using `tap` have had their dependencies installed by a package manager that does not create a flat `node_modules` directory, e.g. [pnpm](http://pnpm.js.org/).

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "opener": "^1.4.1",
     "os-homedir": "^1.0.2",
     "own-or": "^1.0.0",
-    "own-or-env": "^1.0.0",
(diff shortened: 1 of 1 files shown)


jsumners:
I don't see any reason why AppVeyor would be failing.

isaacs:
This doesn't really seem necessary?  `^1.0.0` will already pull in 1.0.1 in a fresh install, and 1.0.1 only fixed a bug that didn't affect tap in the first place (since it also has `own-or` as a dep.)

PR #424, opened 2018-02-16
Fix typos in docs + update contributing guide

customcommander:
Hi

According the the Cambridge dictionary, [hierarchy has been misspelt](https://dictionary.cambridge.org/dictionary/english/hierarchy).

The contributing guide says that docs should be changed on the `gh-pages` branch but that doesn't longer seem to be the case. I took the liberty of updating the guide but happy to take out the offending commit if that's not correct :)


isaacs:
Thanks!

PR #356, opened 2017-03-13
Browserifyable

formula1:
I love browserify, I love node tap, I want them to work togethor in harmony 

The following pull requests also need to be merged
https://github.com/othiym23/only-shallow/pull/6
https://github.com/othiym23/node-deeper/pull/6

PR #427, opened 2018-02-22
Browserifyable 2

markandrus:
Hi, I updated @formula1's PR, #356. I also added a Karma test that uses karma-mocha to test the browserified version of tap. Eventually, I'd like to test the browserified version with an updated karma-tap (or a karma-node-tap), but there are still details to iron out there (e.g., just parse `console.log` like karma-tap, or expose a global TAP/Test instance and hook into its Readable Stream).

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -25,6 +25,7 @@ const Pool = require('yapool')
 const TestPoint = require('./point.js')
 const parseTestArgs = require('./parse-test-args.js')
 const loop = require('function-loop')
+const captureStackTrace = require('capture-stack-trace')
 const path = require('path')
 
 const extraFromError = require('./extra-from-error.js')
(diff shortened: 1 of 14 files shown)


markandrus:
I think Appveyor is failing for reasons unrelated to this PR.

lib/test.js
```diff
@@ -40,6 +41,12 @@ const Promise = require('bluebird')
 const bindObj = require('bind-obj-methods')
 const cwd = process.cwd()
 
+// TODO: Update stack-utils to use capture-stack-trace, too.
```
markandrus, line 44: @isaacs WDYT about updating stack-utils to use capture-stack-trace, too?

isaacs:
Landed this (with some rebasing and updating of dependencies) on v12.1.

isaacs:
Thanks!

PR #429, opened 2018-03-05
docs: add unique page titles

stephenmathieson:
This patch adds unique page titles to the website. When a `title` is present in the page frontmatter, the layout will use it. When a title is not specified, the default `site.title` is used.

This makes navigating through the site much easier, especially when browsing with multiple tabs.


isaacs:
Cool, thanks!

PR #436, opened 2018-04-02
Add twing to the 100 club

ericmorand:
It's been a long ride to get there since TwigPHP code coverage is subpar at best.
It's with great pride that I request Twing to join the 100 club.


isaacs:
Ah, looks like I landed this a while ago.

Thanks!

PR #441, opened 2018-05-02
add dotenv to the tap 100 list

maxbeatty:
https://github.com/motdotla/dotenv/pull/304 now uses tap to continue 💯 coverage


isaacs:
Thanks!

Issue #274, opened 2016-07-15
[Feature] Multiple coverage reports


francisdb:
We would for out ci build like to have multiple coverage reports as `cobertura` is used by jenkins and `lcov` for badge generation

something like `--coverage-report=lcov,cobertura`

isaacs:
This is tricky, but I can see the value in it.  Patch welcome.

Normally I'd prefer an option that is set multiple times rather than a delimited thing, but in this case, you want to be able to override it.  (For example, if your test script is `tap test/*.js --reporter=spec` you might want to do `npm test -- -Rtap` to show the raw TAP output.)  Also, the values are a discrete set of strings, so delimiting is simple and deterministic.

mourner:
Note that `nyc` already supports multiple reporters (through multiple `--reporter` settings). So as a workaround for now, you could do `nyc --repoter=lcov --reporter=cobertura tap --no-coverage test/*.js`.

evanlucas:
We use somewhat of a workaround with npm scripts. In your package.json, you can do something like this:

```
{
  "name": "test-project",
  "version": "1.0.0",
  "scripts": {
    "test": "tap test --cov",
    "posttest": "tap --coverage-report=text-summary"
  }
}
```

This also has the added benefit of being able to change the default reporter from the cli without changing your package.json: 

#### Example

``` bash
$ npm test -- --coverage-report=html
```

This will generate the html coverage report and then the text-summary

cburroughs:
I'd also like to specify multiple reporters by way of `.nycrc`.  The problem I'm running into is that if `--coverage-report` is on the tap argv, then the invocation of nyc with `[nycBin, 'report', '--reporter', options.coverageReport]` takes precedence with what is in `.nycrc`.  If there was some sentinel value (config? empty string?) for `--coverage-report` that caused the invocation to be `[nycBin, 'report']` then multiple reporters by way of the config file would just work.

isaacs:
This was a breaking change, so I got it into v14.  Now you can specify multiple coverage reports like `tap --coverage-report=html --coverage-report=text` and it'll pass them all to nyc.

PR #446, opened 2018-05-17
support multiple coverage reports

cburroughs:
Multiple coverage reports are supported either by using comma separate values on
the command line, or by letting nyc handle the absence of any reporters.  nyc
will either use the reporters in it's own config file, or default to the text
behavior as tap did previously.

ref #274

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -91,7 +91,7 @@ const main = _ => {
 
   // this is only testable by escaping from the covered environment
   /* istanbul ignore next */
-  if ((options.coverageReport || options.checkCoverage) &&
+  if ((options.coverageReporters.length || options.checkCoverage) &&
       options.files.length === 0)
     return runCoverageReport(options)
 
@@ -136,7 +136,7 @@ const constructDefaultArgs = _ => {
     bail: false,
     saveFile: null,
     pipeToService: false,
-    coverageReport: null,
+    coverageReporters: [],
     browser: true,
     coverage: undefined,
     checkCoverage: false,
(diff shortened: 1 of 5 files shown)


cburroughs:
I think I've run down the travis failure as a per-existing issue: #447

cburroughs:
After rebaseing, the travis failures are only on Windows.  They are also failing on master.

isaacs:
If we're going to specify multiples of something, it should be multiple fields, rather than comma-separated.  Take a look at how the `--nyc-arg` and `--node-arg` options are implemented.  Instead of parsing to a string, it's always an array, and each new instead of the option adds its value to the array.

isaacs:
ie, instead of `tap test/*.js --coverage-report=text,html` you'd do `tap test/*.js --coverage-report=text --coverage-report=html`

cburroughs:
I believe the`test-browser` failures are per-existing.  For example https://travis-ci.org/tapjs/node-tap/jobs/490393700

I am not attached to `nyc-default` as sentinel value and struggled to come up with something short to mean "just let NYC take care of it".

isaacs:
Hey, sorry for the delay.

Yes, the browser tests are completely borked, and have been for a while, as far as I can tell.  I intend to yank them out entirely and re-think the whole "run in a browser" thing.  The dep on `write-file-atomic` seems to be completely browser-incompatible, so that's the thing to work out.  And it serves me right for not following my own dogma that tests that don't get run every time are failing by definition.

I've been a bit focused on other things to get v13 completed (https://github.com/tapjs/tap-yaml and https://github.com/tapjs/tcompare, specifically), but I'll be wrapping up some of those soon and can swing back to getting this landed on v12 pretty soon.

It'll have to be implemented a slightly different way for v13, since I'm using a new arg parsing and config approach, but I think it'll actually be easier than this was, especially since you've already figured out how it ought to work.

Issue #335, opened 2016-12-14
Allow Multiple Reporters


shaicantor:
Can you add the feature to add multiple reporters?
For example: spec and xunit.
Meaning, writing spec output and xunit xml file.
There are a few reporters that are doing this, but the tap-mocha-reporter used by tap doesn't allow it.

https://www.npmjs.com/package/spec-xunit-file
https://www.npmjs.com/package/tap-xunit-file

wking:
There's lots and lots of discussion in the analogous Mocha PR: mochajs/mocha#2184.  The main sticking point there seems to be [figuring out how to pass configuration options into multiple reporters][1].

[1]: https://github.com/mochajs/mocha/pull/2184#issuecomment-351556819

isaacs:
The way that tap does reporters is really not very amenable to having multiple of them running at the same time, since a reporter can be an ink React component that takes over the whole terminal.

But, with tap 18, you can run `tap replay --reporter=whatever` to run the test output through a different reporter without actually running all the tests again.  Second process, so no issues figuring out how to specify which reporter gets which options, etc.

Check it out with `npm i tap@pre`. Be sure to read the [changelog](https://node-tap.org/changelog/#18.0) as a lot has changed.

PR #452, opened 2018-06-08
RFC: bin/run: Add --reporter-options

wking:
This PR depends on tapjs/tap-mocha-reporter#49; review that first.

This is inspired by [Mocha's][1]:

```
  --reporter-options <k=v,k2=v2,...>
```

Except:

* I'm using a JSON value for easier parsing and more explicit typing.  This will end up using a few more characters, but the format is more explicit than Mocha's (which only uses string values?).  And callers are likely to already be familiar with JSON, so we save the mental overhead of teaching them a new serialization format.

* I'm using a `reporter` namespace.  This allows us to add other namespaces in the future to address other configurable aspects of tap-mocha-reporter's `Formatter`.  For example, we could use this same CLI option to configure the runner.

Somewhat related to this, if we have plans for allowing multiple reporters (#335), we may want to namespace *those* now.  For example:

```
  --reporter-options {"reporter": {"progress": {"open": "(", "close": ")"}}}
```

to avoid [the issues Mocha bumped into][3] when trying to add multi-reporter support.

[1]: https://mochajs.org/#usage
[2]: https://github.com/tapjs/node-tap/issues/335
[3]: https://github.com/mochajs/mocha/pull/2184#issuecomment-351556819

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -127,6 +127,7 @@ const constructDefaultArgs = _ => {
     timeout: +process.env.TAP_TIMEOUT || defaultTimeout,
     color: !!colorSupport.level,
     reporter: null,
+    reporterOptions: {},
     files: [],
     grep: [],
     grepInvert: false,
@@ -280,6 +281,10 @@ const parseArgs = (args, options) => {
 
       case '--reporter':
         options.reporter = val || args[++i]
+        continue
+
+      case '--reporter-options':
+        options.reporterOptions = JSON.parse(val || args[++i])
         continue
 
       case '--gc': case '-gc': case '--expose-gc':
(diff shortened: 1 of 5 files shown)


isaacs:
There's now in v13 `--reporter-arg=<arg>` or `-r<arg>`which can be set multiple times.  It's passed as arguments to the reporter constructor, if it's a stream module, or set as command-line arguments if it's a CLI program.

Try it out!  `tap -Rtap-parser --reporter-arg=-t --reporter-arg=-f`, and it'll spit out flattened TAP.

I'm more likely to port reports _out_ of tap-mocha-reporter than add more functionality over there, tbh.

PR #456, opened 2018-06-15
Output plan when using CLI with -j option

bas080:
When using the -j option the tap output did not include the plan. This PR tries to fix this issue.

I look forward to getting some feedback on the correctness of this solution.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -753,12 +753,17 @@ const runTests = options => {
 
   // if not -Rtap, then output what the user wants.
   // otherwise just dump to stdout
-  tap.pipe(options.reporter === 'tap' ? process.stdout: makeReporter(options))
+  const outStream = options.reporter === 'tap' ? process.stdout : makeReporter(options)
+
+  tap.pipe(outStream)
 
   // need to replay the first version line, because the previous
   // line will have flushed it out to stdout or the reporter already.
   if (options.outputFile !== null)
     tap.pipe(fs.createWriteStream(options.outputFile)).write('TAP version 13\n')
+
+  if (options.files.length > 1 && options.jobs > 1)
+    outStream.write(`1..${options.files.length}\n`)
 
   saveFails(options, tap)
 
--- a/tap-snapshots/test-run.js-TAP.test.js
+++ b/tap-snapshots/test-run.js-TAP.test.js
@@ -349,6 +349,7 @@ ok 4 - x/y/1.js # {time}
 
 exports[`test/run.js TAP parallel > output 1`] = `
 TAP version 13
+1..5
 ok 1 - p/y/1.js # {time} {
     ok 1 - one
     1..1


isaacs:
It does already output a plan at the end of the output when using buffering for multiple jobs.

Can you please provide a failing test case to demonstrate the problem?  This patch will produce incorrect output (multiple plans), and in fact, can output the wrong plan if one of the command line files encountered is a folder with multiple tests inside of it.

bas080:
You are correct. My proposed patch is incorrect.

There is however something unexpected happening. I found this issue when running my tests using unix's `prove` command. It returns non zero when the tap format is not compliant with TAP requirements. One of the requirements are that tap output should contain the plan.

This is what I am running. 

> ./t/unit.t
```
#!/bin/bash

./node_modules/.bin/tap -j25 ./lib/{**/,}*.spec.js
```

`$ ./t/unit.t | tail`

```
        ok 10000 - expected to not throw
        1..10000
    ok 4 - something # time=688.781ms
    
    1..4
    # time=8345.392ms
}

1..24
# time=11042.014ms
```

So it does contain the plan `1..24`. So what is prove complaining about?

So why does prove exit with non zero?

`$ prove`
```
t/unit.t .. 
All 1 subtests passed 

Test Summary Report
-------------------
t/unit.t (Wstat: 0 Tests: 1 Failed: 0)
  Parse errors: No plan found in TAP output
Files=1, Tests=1,  8 wallclock secs ( 0.26 usr  0.00 sys +  8.70 cusr  0.38 csys =  9.34 CPU)
Result: FAIL
```

I should also say that the tests always pass when running them without the prove test harness. I don't think prove is to blame though.

What I think is happening is that node-tap is having a hard time writing to stdout when there is lots to write and when the process that is reading from stdout is taking its time to read from stdin. Somehow node-tap ends up dropping some of the lines it should be writing to stdout.

I can prove this by using a temp file that will allow me to write the stdout quicker than prove might be processing the stdin.

> ./t/unit.t

```
#!/bin/bash

./node_modules/.bin/tap -j25 ./lib/index.spec.js > a

cat a
```

Somehow this works every time I run the tests using prove.

I am concluding that node-tap does not write the complete TAP output when the process that is consuming the stdout does it too slowly.

This is a completely different issue.

isaacs:
Is it possible that prove isn't properly handling the buffered test output style?

Ie, does it work when you _don't_ use the `-j` argument?

bas080:
@isaacs, when I run the tests using prove but without tap's the -j option, the tests do pass.

The tests also pass when running the tests with the `-j` option and writing the stdout to a temp file.

> does not work most of the time.

```bash
./node_modules/.bin/tap -j25 ./lib/{**/,}*.spec.js
```

> works when using a temp file

```bash
tmp="$(mktemp)"

./node_modules/.bin/tap -j25 ./lib/{**/,}*.spec.js > $tmp

cat $tmp
```

> also works without tap's -j option

```bash
./node_modules/.bin/tap ./lib/{**/,}*.spec.js
```

So it's not that prove does not handle the buffered output incorrectly. I say this because the tests pass when writing the stdout to a temp file before using cat to write it to stdout.

You can try the tests yourself. The repository is here: https://github.com/bas080/sumtin.

You can run `prove` command in the root directory after installing the npm dependencies.

PR #458, opened 2018-06-29
Upgrading node tap deps

jrgleason:


jrgleason:
Humm I see AppVeyor is having trouble, I am using Node 11 not 8 so that could be part but could also be a couple breaking tests that are breaking on master.

```
test/test.js ...................................... 141/141
total ........................................... 5417/5419
  

  5417 passing (25s)
  2 failing
```

isaacs:
All deps already updated.  Also, I unplugged AppVeyor in favor of Travis's Windows testing.

Issue #459, opened 2018-07-03
Earliest compatible node version is not 0.10


ZJONSSON:
The README states that tap `works on any version of Node.js back to v0.10.`   In reality the library now uses `const` and fat arrow functions and does not work for node version prior to `4`.

It might be helpful to update the  README to explicitly state the earliest node version supported (I think it's 4) and possibly provide a link to the latest possible version of tap that  that supports v0.10 and up.

Any library that wants to maintain v0.10 backwards compatibility tests will have to pin tap to an old version.
For example I'm not able to use the latest version [here](https://travis-ci.org/ZJONSSON/node-unzipper/builds/399718363) and will have to revert back to the old tap.

isaacs:
Yes, that is an oversight in the documentation. Node 0.10 is out of LTS, and insecure to continue using. Doc patch welcome, or I’ll fix soon.

isaacs:
d7bd496cb21d080512a059634e68ebd0a05f1f84

ZJONSSON:
Thanks @isaacs !

PR #474, opened 2018-11-09
fix minimum compatible version in docs

ovhemert:
According to the tests, the minimum Node.js version that is supported is v4.x.

Updated the readme for this, and seems to close an issue as well. #459


isaacs:
Thanks!

PR #481, opened 2018-11-24
Add easy and straightforward ts-node support

ilyadoroshin:
Long story short, it was really upsetting trying to setup tap to work with typescript. 

I've had a lot of difficulties running standalone test files with VSCode launch.json due to the fact that all my .test.ts needed shebang. It was a terrible mess.

Please find PR for new option --ts, --typescript that would run `node -r ts-node/register` on all your .ts files. 

Example of usage:
```
TS_NODE_PROJECT=./tsconfig.test.json ./node_modules/.bin/tap --reporter=spec --ts tests/**/*.test.ts
```

Works like charm! And I can debug my test files with the following launch.json:
```
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Current TS Tests File",
      "type": "node",
      "request": "launch",
      "protocol": "inspector",
      "env": {
        "TS_NODE_PROJECT": "tsconfig.test.json"
      },
      "args": ["${relativeFile}"],
      "runtimeArgs": ["-r", "ts-node/register"],
      "sourceMaps": true,
      "cwd": "${workspaceRoot}"
    }
  ]
}
```


ilyadoroshin:
and here's [my tap.d.ts](https://gist.github.com/ilyadoroshin/eb02ebf83b8ccf852830454508f5522f) just in case someone needs it

ilyadoroshin:
doesn't work with `--coverage` flag, fixing

ilyadoroshin:
Looks like TypeStrong/ts-node#564 fixes the issue and you can run tests with coverage checks, however nyc wouldn't show you coverage in terminal for some reason

ilyadoroshin:
https://github.com/ilyadoroshin/node-tap-ts-example

huan:
Good job!

test/typescript.js
```diff
@@ -0,0 +1,47 @@
+'use strict'
+var t = require('../')
```
isaacs: These vars should all be either `const`s or `let`s.

bin/run.js
```diff
@@ -507,6 +513,11 @@ const runCoverageReportOnly = (options, code, signal) => {
 
   const args = [nycBin, 'report', '--reporter', options.coverageReport]
 
+  if (options.typescript) {
+    args.push('--reporter');
+    args.push('text');
+  }
```
isaacs, line 519: I don't understand this.  Why are you forcing coverage with a text reporter when `--type-script` is enabled?

test/typescript.js
```diff
@@ -0,0 +1,47 @@
...
+  try { fs.unlinkSync(ts) } catch (e) {}
+  try { fs.rmdirSync(fixdir) } catch (e) {}
+}
+
+t.test('setup', function (t) {
+  cleanup()
+  fs.mkdirSync(fixdir, '0755')
+  fs.writeFileSync(
+    ts,
+    'import * as tap from \'../../\';\n' +
```
isaacs: Style nit: prefer using `` ` `` for multiline strings.  Ie, instead of this, use something like:

```js
fs.writeFileSync(ts, `
import * as tap from '../../';
tap.test('TS Works', (t) => {
  t.equals(2, 1 + 1);
  t.end();
});
`)
```

isaacs (commented):
This looks good, but for a few relatively minor comments and style nits.

bin/run.js
ilyadoroshin: @isaacs for some reason, when --ts is on, nyc doesn't show coverage in terminal but builds coverage into specified folder. It's just discrepancy that I've found

ilyadoroshin:
@isaacs all fixed

package.json
```diff
@@ -81,6 +81,8 @@
     "karma-spec-reporter": "0.0.32",
     "mocha": "^5.0.1",
     "standard": "^12.0.1",
+    "ts-node": "^7.0.1",
+    "typescript": "^3.1.6",
```
marvinhagemeister: Suggestion: If this would be declared as a `peerDependency` users could chose the TS version they prefer. This would prevent us from having to update it ourselves whenever a new TS version comes out.

andykais:
would you be willing to include your tap.d.ts file in this PR? https://github.com/Microsoft/dtslint could be used to perform some simple tests against the methods.

ilyadoroshin:
@andykais you want to make tap.d.ts a part of the PR and to check it wether it works correctly? It just sounds like not part of the scope, I would rather create it on for @types/node-tap and perform linting there

andykais:
@ilyadoroshin fair enough, I suppose definition files only make sense in repos that are using typescript. I figured it was worth asking. I dont feel comfortable enough with this library (or modifying definition files) to PR something against [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped)

ilyadoroshin:
@andykais one thing at a time, man :) i do believe there's no `@types/node-tap` but it would be awesome if someone could help me with it, since i'm a bit new to dtslint. anyways, good thinking, we can at least try to get awareness of community so that everyone could collaborate to make good typings for this awesome lib

bin/usage.txt
```diff
@@ -92,6 +92,11 @@ Options:
                               captured by default and sent to the
                               coveralls.io service.
 
+  --typescript --ts           Run .ts files with `node -r ts-node/register`
+                              
+                              In order for this to work,
+                              you must intsall `ts-node` locally.
```
joeledwards, line 98: Typo: `intsall` -> `install`

Also, is a local install necessary given that `ts-node` was added to `package.json`?

joeledwards: Ah, missed that it was a dev. dep.

ilyadoroshin: there's a good thought to move ts-node to peerDeps, will do today

isaacs:
I decided to do this an even more easy and straightforward way.

If you tell tap to run a `.ts` file, it'll use `ts-node` to do it.  No extra args or peer deps necessary.

If you really want to use your own typescript for whatever reason, you can do that with `--node-args` like you were before, so the workaround still exists.

isaacs:
That just landed on 12.4, btw.  697fe60

PR #483, opened 2018-12-05
Move slash replacement to the end

cemremengu:
Make sure the slash replacement is done

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -975,10 +975,11 @@ class Test extends Base {
 
   get fullname () {
     const main = process.argv[1] || ''
-    return (this.parent ? this.parent.fullname
+    const name = (this.parent ? this.parent.fullname
       : main.indexOf(cwd) === 0 ? main.substr(cwd.length + 1)
-      : path.basename(main)).replace(/\\/g, '/') +
-      ' ' + (this.name || '').trim()
+      : path.basename(main)) + ' ' + (this.name || '').trim()
+    
+    return name.replace(/\\/g, '/')
   }
 
   matchSnapshot (found, message, extra) {
--- a/lib/test.js
+++ b/lib/test.js
@@ -975,11 +975,9 @@ class Test extends Base {
 
   get fullname () {
     const main = process.argv[1] || ''
-    const name = (this.parent ? this.parent.fullname
-      : main.indexOf(cwd) === 0 ? main.substr(cwd.length + 1)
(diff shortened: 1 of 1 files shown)


cemremengu:
Just tested this manually from my `node_modules` and it fixes the problem. I was going to add a test case but I think it is best to enable windows testing on travis for these kind of things once it is ready.

I would buy you guys a coffee if you can patch this soon :)

lib/test.js
```diff
@@ -975,10 +975,11 @@ class Test extends Base {
...
   get fullname () {
     const main = process.argv[1] || ''
-    return (this.parent ? this.parent.fullname
+    const name = (this.parent ? this.parent.fullname
       : main.indexOf(cwd) === 0 ? main.substr(cwd.length + 1)
-      : path.basename(main)).replace(/\\/g, '/') +
-      ' ' + (this.name || '').trim()
+      : path.basename(main)) + ' ' + (this.name || '').trim()
+    
+    return name.replace(/\\/g, '/')
```
isaacs: Won't this be an issue if you _want_ to have a slash in the name?

For example: `t.test('make sure \\ is replaced with /', t => { ... })` would result in some confusing output.

cemremengu: Oh, yes you are right. Not sure why the current one doesnt work even though the replacement. Whats a more suitable place to handle this? Any hints are appreciated.

isaacs: Yeah, I'm guessing that maybe the name is getting set somewhere in the `Spawn` subclass perhaps?  I'd been using the Windows support on Travis briefly, but ran into too many issues around signal and subprocess differences, so I think the only way forward is for me (or someone) to spin up a Windows VM and get all the tests passing locally first.  The turnaround is just too slow to do with a remote CI system otherwise, and a known-failing test is worse than no test at all.

cemremengu:
I am confused actually. The version I have (^12.0.0) has a different `fullname` method than the master. The version in master works fine actually. Maybe I should have more coffee not sure what's going on 😄 

Not sure why semver is not bumping to next minor version which this was fixed

isaacs:
Oh!  Looks like I never published the code on master branch, because I was still fantasizing that I'd get everything passing on Windows first.  Maybe I should just bite the bullet and accept it as-is :)

cemremengu:
Since I use windows, I can try getting tests to run locally first. Not sure how tough it will be for me though, this project looks beyond my skills but I am up for the challange :)

PR #485, opened 2018-12-06
Add attributes file to force LF

cemremengu:


cemremengu:
Landed in 04d27df211d52831459810d5b7be46d311bbaf97

PR #488, opened 2018-12-18
Add 'identical' to synonyms

troublete:
This adds 'identical' as synonym to equal. To have a certain symmetry to same's 'equivalent'.


isaacs:
699b179 Thanks

PR #505, opened 2019-01-30
Snapshots: use pretty-format instead of util.inspect

isaacs:
`util.inspect` has sometimes changes between Node.js versions, making cross-version testing challenging.  This is the snapshot formatter that Jest is using, it seems pretty good and well-maintained.

This is a breaking change, of course, because everyone will have to re-generate their snapshots.


isaacs:
Merged into the v13 branch, it'll go out in that release.

PR #507, opened 2019-02-08
Don't emit end while waiting for pipes to clear

isaacs:
Slight tweak on #506 to avoid making a breaking change in other use cases.

@kalinkrustev can you please check to see if this fixes your issue with using `--output-file`?

Thank you very much for finding this and putting up with my pedantry around the fix :)

### DIFF
--- a/lib/base.js
+++ b/lib/base.js
@@ -182,7 +182,8 @@ class Base extends MiniPass {
       this.options.failures = failures
 
     this.onbeforeend()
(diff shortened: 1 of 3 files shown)


kalinkrustev:
You are welcome. I also thank you for quickly picking and finishing the fix properly.
I think this should fix the issue, I tested with my simple cases, it worked.
I will report if my original big test fails.

PR #521, opened 2019-03-22
Fix font-family typo: Part II

jordanbtucker:
Fixes monospace fonts on platforms that don't have any of the named fonts, like Chromebooks. Like #520, but for code blocks.


isaacs:
Landed on 01c0f652d2f54c57dbbe16249fb8c45760c4face (v13) and 775e8545d3baf3d56fa149d1779ae0a0ec91abf5 (v12)

Issue #522, opened 2019-03-22
TypesScript support on windows not working


fox1t:
Hi!
Testing TS code is not working at the moment. Current implementation uses [ts-node bin](https://github.com/TypeStrong/ts-node/blob/master/package.json#L7) field that is [not cross-platform](https://github.com/TypeStrong/ts-node/blob/master/src/bin.ts#L1).
The right way to use ts-node cross-platform is to pass `-r ts-node/register` option to node, instead of spawning ts-node child process.

I'll make a PR to fix this specific issue.

This is the full error dump on Windows 10:

```
 TAP version 13
not ok 1 - ./index.ts # time=20.695ms
  ---
  timeout: 30000
  file: ./index.ts
  childId: 0
  command: >-
    C:\temp\node_modules\ts-node\dist\bin.js
  args:
    - '-r'
    - >-
      C:\temp\node_modules\esm\esm.js
    - ./index.ts
  stdio:
    - 0
    - pipe
    - 2
  cwd: 'C:\temp'
  ...
{

    not ok 1 - spawn UNKNOWN
      ---
      timeout: 30000
      file: ./index.ts
      childId: 0
      command: >-
        C:\temp\node_modules\ts-node\dist\bin.js
      args:
        - '-r'
        - >-
          C:\temp\node_modules\esm\esm.js
        - ./index.ts
      stack: |
        loop (node_modules/function-loop/index.js:35:15)
      stdio:
        - 0
        - pipe
        - 2
      cwd: 'C:\temp'
      at:
        line: 366
        column: 11
        file: internal/child_process.js
        function: ChildProcess.spawn
      errno: UNKNOWN
      code: UNKNOWN
      syscall: spawn
      test: ./index.ts
      ...
    1..1
    # failed 1 test
}

1..1
# failed 1 test
# time=44.501ms
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! temp@1.0.0 test: `tap -J ./index.ts`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the temp@1.0.0 test script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\Baz\AppData\Roaming\npm-cache\_logs\2019-03-22T11_30_27_215Z-debug.log 
```

fox1t:
I opened [523](https://github.com/tapjs/node-tap/pull/523) PR.

PR #523, opened 2019-03-22
Passes `-r ts-node/register` argument to node

fox1t:
Fixes cross-plaftorm issues.
We can safely use node `-r ts-node/register` argument since ts-node and TypeScript are tap dependencies.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -16,9 +16,6 @@ const path = require('path')
 const exists = require('fs-exists-cached').sync
 const os = require('os')
 const isTTY = process.stdin.isTTY || process.env._TAP_IS_TTY === '1'
-const tsNode = require.resolve(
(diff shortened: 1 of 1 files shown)


fox1t:
I opened [523](https://github.com/tapjs/node-tap/pull/523) PR.

fox1t:
Relates to [522](https://github.com/tapjs/node-tap/issues/522)

isaacs:
This works for plain typescript, but is there a way to pass compiler options to ts-node when using `-r`?  On the v13 branch, I'm doing `ts-node --compiler-options='{"jsx":"react"}' ok.tsx` for tsx files.

isaacs:
Ah, nevermind I figured it out.  The answer is the `TS_NODE_COMPILER_OPTIONS` environment variable.

PR #534, opened 2019-04-20
Update to latest tap-mocha-reporter

adroitwhiz:
This should fix _some_ of the pesky js-yaml security advisories. While it was a major version bump, there weren't any breaking changes AFAIK.


isaacs:
This update is a breaking change because of the change to the parser, and swapping out js-yaml with tap-yaml.

Tap v13 doesn't have the problem, and I believe that js-yaml@3.13.1 fixes the issue, so you should be able to just `npm update` and it'll fix the problem, or `npm i tap@13` to get the new version and features.

Issue #549, opened 2019-05-03
In a project with mixed .ts and .js tests, .ts is double-compiled in tap 12.6.1+


kyegupov:
I have followed the advice given in https://github.com/tapjs/node-tap/issues/526 and added `-r ts-node/register` flag.

This behavior got broken in commit https://github.com/tapjs/node-tap/commit/cce3d7215a1d949e4ba88e64db5d816d29e3ba33

Now, for .ts files, tap will add ts-node/register twice (once by itself, once supplied by me). This results in .ts files being transpiled twice (and the second transpile usually fails due to type annotations being stripped by the first one).

This "guessing" behaviour causes only trouble. Can we just nave a flag for "always use ts-node"?

Issue #526, opened 2019-04-02
Typescript support broken for mixed .ts/.js projects


kyegupov:
Recently, Tap has introduced support for running `.ts` files, like this:

https://github.com/tapjs/node-tap/blob/1871736bfaf39dd7b1be8de4b9359e2eeaa1e4ef/bin/run.js#L748-L753

However, imagine a project being slowly migrated from JS to TS. There are some `.js` tests that would refer to `.ts` code. But, see the code above: ts-node won't be used for `.js` tests and it will not be able to load `.ts` modules they reference.

The proper way to solve it would be to either always use ts-node (if available) to run files. Or, at least, to have some kind of `--force-ts-node` flag.

isaacs:
Try this: `tap --node-arg=-r --node-arg=ts-node/register`

In version 13, you can put this in your package.json, even:

```json
{
  "tap": {
    "node-arg": ["-r", "ts-node/register"]
  }
}
```

isaacs:
You can also add `require('ts-node/register')` to the top of all your js files, which would make them able to require() a ts file.

PR #550, opened 2019-05-03
Add --ts flag to properly support projects with mixed .js and .ts files.

kyegupov:
Properly addresses issues https://github.com/tapjs/node-tap/issues/549 and https://github.com/tapjs/node-tap/issues/526

### DIFF
--- a/bin/jack.js
+++ b/bin/jack.js
@@ -544,9 +544,13 @@ Much more documentation available at: https://www.node-tap.org/
   }),
 
   esm: flag({
-    default: process.env.TAP_NO_ESM !== '1',
     description: `Run .js and .mjs with support for EcmaScript modules
                   (Default: true)`,
+  }),
+
+  ts: flag({
+    description: `Always use Typescript loader. Useful when .js tests
+                  require .ts code`,
   }),
 
   'nyc-help': flag({
--- a/bin/run.js
+++ b/bin/run.js
@@ -577,29 +577,24 @@ const runAllFiles = (options, tap) => {
       if (options.jobs > 1)
         opt.buffered = isParallelOk(parallelOk, file) !== false
 
-      if (file.match(/\.m?js$/)) {
-        const args = [
+      if (file.match(/\.m?js$/) || file.match(/\.tsx?$/)) {
+        let args = [
           ...(options.esm ? ['-r', esm] : []),
           ...options['node-arg'],
           file,
           ...options['test-arg']
         ]
-        tap.spawn(node, args, opt, file)
-      } else if (file.match(/\.tsx?$/)) {
-        const compilerOpts = JSON.stringify({
-          ...JSON.parse(process.env.TS_NODE_COMPILER_OPTIONS || '{}'),
-          jsx: 'react'
-        })
-        opt.env = {
-          ...env,
-          TS_NODE_COMPILER_OPTIONS: compilerOpts,
+        if (options.ts || file.match(/\.tsx?$/)) {
+          const compilerOpts = JSON.stringify({
+            ...JSON.parse(process.env.TS_NODE_COMPILER_OPTIONS || '{}'),
+            jsx: 'react'
+          })
+          opt.env = {
+            ...env,
+            TS_NODE_COMPILER_OPTIONS: compilerOpts,
+          }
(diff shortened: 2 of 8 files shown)


bin/jack.js
```diff
@@ -544,11 +544,15 @@ Much more documentation available at: https://www.node-tap.org/
   }),
 
   esm: flag({
-    default: process.env.TAP_NO_ESM !== '1',
```
isaacs, line 547: This seems unrelated?

isaacs (commented):
I'm curious why we can't just detect the double-instrumentation and prevent it?  The problem seems like it arises when you have something like this:

```js
// load-ts.js
require('./ok.ts')
```

```typescript
// ok.ts
import * as t from './'
t.pass('this is fine')
```

```
node -r ts-node/register -r ts-node/register load-ts.js
```

But when I run that, it seems like it works fine, so I'm a little bit confused about whether I understand the problem this is trying to address.  Do you have an example where I can reproduce the problem you're experiencing?

Also: note that this PR is to the master branch, which is version 13.  To address the issue in a v12 legacy release, it'll have to be done (a little bit differently) in the `v12` branch.  Maybe not an issue you care about if you want to upgrade anyway, but worth mentioning.

isaacs:
Oh!  Maybe what you're describing only happens when there are two _different_ instances of ts-node in play?  I copied ts-node to another folder, and ran it this way.  This is an interesting problem:

```
$ node -r ts-node-duplicate/register -r ts-node/register load-ts.js

/Users/isaacs/dev/js/tap/node_modules/ts-node/src/index.ts:240
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
ok.ts:2:23 - error TS2304: Cannot find name 'exports'.

2 Object.defineProperty(exports, "__esModule", { value: true });
                        ~~~~~~~
ok.ts:3:9 - error TS2580: Cannot find name 'require'. Do you need to install type definitions for node? Try `npm i @types/node`.

3 var t = require("./");
          ~~~~~~~

    at createTSError (/Users/isaacs/dev/js/tap/node_modules/ts-node/src/index.ts:240:12)
    at reportTSError (/Users/isaacs/dev/js/tap/node_modules/ts-node/src/index.ts:244:19)
    at getOutput (/Users/isaacs/dev/js/tap/node_modules/ts-node/src/index.ts:360:34)
    at Object.compile (/Users/isaacs/dev/js/tap/node_modules/ts-node/src/index.ts:393:11)
    at Module.m._compile (/Users/isaacs/dev/js/tap/node_modules/ts-node/src/index.ts:439:43)
    at Module.m._compile (/Users/isaacs/dev/js/tap/node_modules/ts-node-duplicate/src/index.ts:439:23)
    at Module._extensions..js (internal/modules/cjs/loader.js:770:10)
    at require.extensions.<computed> (/Users/isaacs/dev/js/tap/node_modules/ts-node-duplicate/src/index.ts:442:12)
    at Object.require.extensions.<computed> [as .ts] (/Users/isaacs/dev/js/tap/node_modules/ts-node/src/index.ts:442:12)
    at Module.load (internal/modules/cjs/loader.js:628:32)
```

isaacs:
Interestingly, if I do what the error message says, and `npm i @types/node`, it fixes the problem in that case as well.  Of course, not as efficient to double-transpile it since the second one has nothing to do, but still, not a crash.

isaacs:
It seems like a cleaner way to address this would be to treat `--ts` or `--no-ts` in the same way as the `--esm` flag.  Ie, it's a flag that just means "do `-r ts-node/register` for any `/\.tsx?$/` files".  If you want to provide your own mechanism, you'd do `tap --test-arg=--require=my-ts-node/register --no-ts test/foo.ts`.  To make it even easier, you could put this in your package.json:

```json
{
  "tap": {
    "ts": false,
    "test-arg": [
      "-r",
      "my-ts-node/register"
    ]
  }
}
```

Would that address the problem?

isaacs:
@kyegupov Can you check out https://github.com/tapjs/node-tap/commit/9fd8a12ba4fcc0396b383577078f7b6d5735b2c8 and see if that would give you what you need?

PR #560, opened 2019-05-23
Remove sentence about Node 4

mcollina:
I do not think the sentence about Node 4 is useful, as this requires Node 8 which ships promises.


isaacs:
Oh yeah, that's an oversight from long ago for sure.

Issue #561, opened 2019-05-27
Support for flow


newtang:
It'll be handy to support [flow](https://flow.org/), a static type checker for Javascript.

PR #562, opened 2019-05-27
Support for flow #561

newtang:


bin/run.js
```diff
@@ -602,6 +604,17 @@ const runAllFiles = (options, tap) => {
...
 
+      if(options.flow){
+        debug('flow', file)
+        const args = [
+          ...(options['node-arg']),
+          flowNode,
+          file,
+          ...(options['test-arg']),
+        ]
+        tap.spawn(node, args, opt, file)
```
isaacs: This will double-spawn in most cases, I think.  (Ie, here, and then down below when it detects that it's a *.js file.)

Rather than spawn here, I think you can just `options['node-arg'].push('-r', flowNode)` if `--flow` is set, and then let the spawn be handled below.

bin/run.js
```diff
@@ -602,6 +604,17 @@ const runAllFiles = (options, tap) => {
       if (options.jobs > 1)
         opt.buffered = isParallelOk(parallelOk, file) !== false
 
+      if(options.flow){
```
isaacs: Nit: space after the `if` please :)

test/run/flow.js
```diff
@@ -0,0 +1,65 @@
...
+  
+  run(args, {}, (er, o, e) => {
+    t.equal(er, null)
+    t.matchSnapshot(o)
+    t.end()
+  })
+})
+
+
+// t.test('flow manually', t => {
```
isaacs: Remove rather than commenting out.  (We can always pull out of git if it turns out to be necessary, but it usually doesn't.)

tap-snapshots/test-run-flow.js-TAP.test.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+'use strict'
+exports[`test/run/flow.js TAP flow > must match snapshot 1`] = `
+TAP version 13
+ok 1 - cli-tests/flow/ok.js # SKIP no tests found {
+    1..0 # no tests found
+}
+
+not ok 2 - cli-tests/flow/ok.js # SKIP no tests found {
+    1..0 # no tests found
```
isaacs: I think this is the double-spawn I mentioned above.  Note that the test runs twice!

It seems weird that it's saying `1..0 # no tests found`.  Shouldn't it still run the test file?

isaacs (changes requested):
Almost there!  Some changes requested.

test/run/flow.js
newtang: Sorry about that. Sloppy.

tap-snapshots/test-run-flow.js-TAP.test.js
newtang: Fixed

bin/run.js
newtang: 👍

bin/run.js
newtang: I see how that works now. Sorry about that.

newtang:
@isaacs Thanks for your patience! I made your requested fixes.

isaacs:
Thanks!  (And no need to apologize for code not already being perfect, this is a team sport :)

Landed in 14.2.

package.json
```diff
@@ -26,6 +26,7 @@
     "domain-browser": "^1.2.0",
     "esm": "^3.2.25",
     "findit": "^2.0.0",
+    "flow-remove-types": "^2.99.0",
```
noahehall, line 29: havent seen that trick before, nice ;)

PR #582, opened 2019-07-22
Typings + runtime checks

Eoksni:
Adds Typescript definition file (incomplete) and runtime test to validate those types with real world.

Validation is done with `typescript-is` and currently includes:
- [x] "smoke" test to verify the shape of the api
- [ ] "smoke" test to verify no excess properties/functions on the api (can be done easily, not enabled yet because of incomplete typings)
- [ ] return value types of common functions
- [ ] argument types for common functions ? not sure if it makes any sense


Eoksni:
To summarise, right now it checks if typings define something that isn't there.
For example, if you remove `t` from the top-level export, it should catch that.

@isaacs It seems failing travis is not related to this PR. Could you check and also let me know if you are fine with additional dependencies this PR brings?

tap.d.ts
```diff
@@ -0,0 +1,30 @@
+// Type definitions for tap 14.4.2
```
isaacs, line 1: Not a great idea to hard-code the version in there.  It's already out of date ;)

tap.d.ts
```diff
@@ -0,0 +1,30 @@
...
+// Project: tap
+// Definitions by: Dmitry Mazurok <https://github.com/Eoksni>
```
isaacs, line 3: If this lives in the node-tap project, then it's going to be maintained by this project's maintainers.  I generally avoid personal attributions in files like this, because it sets up a weird expectation, and duplicates information already found in the git history.

If you'd prefer to have this work attached to your personal account for reputation or bragging rights, I totally understand.  In that case, it should be a separate project that you maintain.  Contributions to a project should be contributed to the project.

Eoksni: No worries, I'll remove it. It is just a leftover from the template.

tap.d.ts
Eoksni: Yeah, I guess it doesn't make sense to specify version if it ships with the package, especially with the tests they are supposed to be up-to-date ;) It is just a normal practice to hard-code version when typings are external so it was there from the template.

tap.d.ts
isaacs: Oh!  Ok, that makes total sense then.  (Same with the version number.)

isaacs:
Travis failing should be fixed.  (They added a new deprecation warning in node 12, it seems.)

Yeah, I'm totally fine with adding a dev dep to run the tests.

PR #594, opened 2019-10-09
feat: Upgrade to nyc 15.0.0-beta.0

coreyfarrell:
This works for node.js <= 12.1.0.  In 12.2.0+ the MJS tests fail.  This
is likely because nyc patches `require.extensions` before the `esm`
module.  This could be worked around by telling nyc to require `esm`
instead of doing it directly, though that would have the side-effect of
propagating the require of esm (not sure if that is good or bad).

A patch is needed for `append-transform` to suppress `ERR_REQUIRE_ESM`
though I haven't yet determined the best way to accomplish this.  I'd
love to drop `append-transform` in favor of `pirates` though that module
also suffers from `ERR_REQUIRE_ESM`.

---

@isaacs not ready to merge yet but I wanted to post so we could see the results on all CI targets.


coreyfarrell:
@isaacs what do you want to do about `esm` in node.js 12.2.0+, I think I'm nearly ready to release nyc 15.  If `esm` is enabled in tap do you want it to only load for direct children or do you want it in grandchildren?  IE if a test does `child_process.spawn(require.resolve('../lib/bin.js'))` should `esm` be loaded to that process.

coreyfarrell:
Will rebase and resubmit once #631 is merged.

PR #601, opened 2019-10-18
feat: Add JSON Schema for .taprc

pke:
This could also be auto-generated using the definitions from jack.js in a buildstep.

It could also be submitted in a separate PR at `https://github.com/schemastore/schemastore/` for publishing under `http://schemastore.org/taprc`

It works in VS Code together with the [Redhat YAML Extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml)

and the following (workspace) settings (for this repo):

```json
"yaml.schemas": {
  "./.taprc.schema.json": ".taprc"
},
"files.associations": {
    ".taprc": "yaml",
}
```

When its deployed to schemastore it might work automagically. Or it would have to be configured in the settings with either the `raw` path to this repos JSON file or with the `node_modules/node-tap/.taprc.schema.json`

## Things to check before merging this

Please check those entries for validity:

- coverage-report: []
- files: []
- grep: []
- node-arg: []
- nyc-arg: []
- reporter-arg: []
- test-arg: []
- test-env: []

Can they really take either a string or do they always have to be an array?

### DIFF
--- a/bin/jack.js
+++ b/bin/jack.js
@@ -1,4 +1,4 @@
-const { jack, num, opt, list, flag, env } = require('jackspeak')
+const { jack, parse, num, opt, list, flag, env } = require('jackspeak')
 
 const {cpus} = require('os')
 const colorSupport = require('color-support')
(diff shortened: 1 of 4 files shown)


isaacs:
The entry in `taprc` will always be an array, but it's an array of strings.

Since this is a duplication of definitions in `bin/jack.js`, I'd rather prefer that it be generated programmatically from that source.

I notice that it's not in the _package_ in this PR, only in the git repo.  If you intend to pull the schema out of the package, I'd prefer that it be built by a `prepare` script, and ignored by git.  (Ie, add it to `package.json#files`, and not to `.gitignore`.)

My reason for this feedback is that I probably won't use it, so I need a way to create it that doesn't rely on me keeping it up to date, since I'll forget.

pke:
I am all for automation. I'll implement that today, rebase and update this PR.

pke:
@isaacs `jackspeak` seems to have no way of handing me back the parsed options so I could print out the name and description of the option. It does define the `parse_` function which would do that, but that is not exported from `jackspeak`. Any idea how to centralise the options?

I wanted to put a new `--dump-config-schema` option in and with breacking the current jack.js export attach the parsed options to the `main` function:

```js
module.exports = main => {
  main.jackOptions = parse(...jackSections)
  const [mainSection, ...remainingSections] = jackSections
  return jack({ main, ...mainSection }, ...remainingSections)
}
```

And then just create the schema file from within `run.js/main()` using `main.jackOptions`.

Of course I could also just create named exports from `jack.js` that expose the options to `run.js`. But that would break the `jack.js` API and I am not sure if any other tool is using it.

pke:
OK @isaacs, I have updated the PR with an autogenerated file.

It's not complete though. The schema should be accessible at the URI described in the schema itself, ideally `https://node-tap.org/taprc.schema.json`.
When I create a `prepare` step to generate the schema, where do I have to save it for it to be deployed to the root of the website with that name?

isaacs:
I believe it should end up in docs/static.

I don't think `--dump-config-schema` should be a config option, though.  Just create a script that does it, and then do it in the `prepare` step so it gets included in git and pushed to the website.

isaacs:
Also, maybe this is a dumb question, but `.taprc` is actually yaml, not strictly json.  I know json is a subset of yaml, so a json file there _should_ Just Work, but I'm wondering if we're gonna get spurious "invalid json" warnings in some cases if they load this schema file.

pke:
it works flawlessly with YAML files. No worries ;) Check out the redhat [extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml)

Ok, if its not a command line option then I would need to either change the jack export from jack.js and export the options as named exports, or import from a new file called jackOptions (or similar).

What are your preferences?

isaacs:
I think exporting the options as a named export is fine.  `bin/jack.js` isn't exported by default from tap, it's just in the bin folder since it's tested along with the other cli tests, so it makes the coverage-map easy.

pke:
@isaacs the `prepare` step is supposed to be run by any of the netlify steps? Cause I get 404 on the deployed preview site `https://deploy-preview-601--node-tap.netlify.com/taprc.schema.json`

pke:
btw "netlify" is hideous in its usage of scripts and tracking. I can't get it to show any of the details pages on my Firefox (default installation), Chrome (with ScriptSafe and uMatrix) nor IE11. So I can't check which files where [deployed](https://app.netlify.com/sites/node-tap/deploys/5db3961926859e0008ac8a80)

isaacs:
I think the deploy preview you want is https://deploy-preview-601--node-tap.netlify.com/

The link in your comment is the netlify app itself.  I don't know what kind of tracking scripts are present there, but the node-tap.org site itself doesn't have any junk you should be worried about.

bin/dump-taprc-schema.js
```diff
@@ -0,0 +1,95 @@
+const { options, reporters } = require("./jack.js")
```
isaacs: This file probably better belongs in `scripts`, since it's a standalone script, not part of the `bin` unit.

package.json
```diff
@@ -70,6 +70,7 @@
     "posttest": "rm -rf cli-tests-*",
     "postunit": "npm run posttest",
     "t": "node bin/run.js test -J -sfails.txt",
+    "prepare": "node bin/dump-taprc-schema.js > docs/static/taprc.schema.json",
```
isaacs: This isn't going to get run on docs generation, because `gatsby` is executed in the docs folder.

It would probably make the most sense in `docs/package.json` as a `prebuild` command.

isaacs (changes requested):
Minor nit, but I would like to see this actually work and put the file on the website before landing and shipping it.

Another question, how is this file discovered?  Should it have a docs change as well telling people how to include it in their system somehow?

bin/dump-taprc-schema.js
pke: > I think the deploy preview you want is [deploy-preview-601--node-tap.netlify.com](https://deploy-preview-601--node-tap.netlify.com/)


I meant those links in the Checks section of this PR
https://app.netlify.com/sites/node-tap/deploys/5db3961926859e0008ac8a80

They only for a brief moment show something and then the whole site goes black.

package.json
pke: Ok I will change that.

pke:
So @isaacs I've tackled all the issues and it should be good now.

However I can't seem to access `https://deploy-preview-601--node-tap.netlify.com/taprc.schema.json` though. Locally the file is copied to the `public` folder. Maybe gatsby needs to be instructed to serve it (preferably with the correct content-type)? btw why do you use gatsby? Isn't the site all static anyway?

About making this public, I think a note in the readme would be sufficient as I'll also create a PR to [schemastore](https://github.com/SchemaStore/schemastore) once you approve&merge this PR here. This will then make the schema known to editors like VSCode and others (mentioned in the schemastore website) so all users will instantly benefit once the create & edit the .taprc file. Magic at work :)

PR #602, opened 2019-10-18
feat: Remove rcfile from --dump-config

pke:
The `--dump-config` command also dumped the `rcfile` command also which would not make sense to have in the .taprc file, wouldn't it? It would create a kind of recursion or twilight zone?

Instead we print the path to the `.taprc` now as first line of the dump.

This also fixes the dump-config tests for Windows.

Since we now print the path to the `.taprc` file on the console, on Windows this path contained backslashes.
They are sanitized in the `clean` function.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -149,8 +149,13 @@ const main = async options => {
     options.reporter = options.color ? 'base' : 'tap'
 
   if (options['dump-config']) {
-    console.log(yaml.stringify(Object.keys(options).filter(k =>
-      k !== 'dump-config' && k !== '_' && !/^[A-Z_]+$/.test(k)
+    delete options['dump-config']
+    const rcfile = options['rcfile']
+    delete options['rcfile']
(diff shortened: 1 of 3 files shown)


pke:
Any thoughts on that @isaacs?

pke:
Rebasing this gave me like 30 failing tests now. Please let me know if you consider merging this PR (its already 2 months old without any comment) or I can save my time to investigate those failings.

isaacs:
Sorry for the delay.

> The --dump-config command also dumped the rcfile command also which would not make sense to have in the .taprc file, wouldn't it? It would create a kind of recursion or twilight zone?

So this made sense to me at first, but what if you're running `tap --dump-config` specifically to see what rcfile you're using?  It seems a bit weird to dump `tap --dump-config` directly into an rcfile without at least checking to make sure you'er not just saving some defaults, no?

pke:
It's OK. I've moved to jest for all of my tests in the meantime.

PR #607, opened 2019-10-27
Added type definitions

### LINKED ISSUE #237: Creating TypeScript types definition

delvedor:
Hello! This pr adds support to the type definitions of the public API of `tap`.
It's based on https://github.com/tapjs/node-tap/pull/256 and fixes https://github.com/tapjs/node-tap/issues/237.

### How does this work

I know that the requirement for having the type definitions shipped with tap is *"no test, no types"*, and I couldn't agree more.
Unfortunately, as you know, currently it's not possible to test if the type definitions written by hand are in sync with the javascript codebase without writing the whole codebase in typescript.

The solution proposed in this pr is to use [`tsd`](https://github.com/SamVerschueren/tsd), a fantastic library that allows testing the type definitions. We should intend this as a unit test for the type definitions rather than an integration test with the javascript code, which means that if you change the javascript codebase without updating the type definitions, that types test will still pass.

### Why here and not on DefinitelyTyped?

`tap` is a batteries included library, and the need to go to DefinitelyTyped for having the types is not ideal *(same reason why we are not installing `nyc`, for example)*.

Maybe in the future, this situation will improve *(see https://github.com/SamVerschueren/tsd/issues/46)* and we will be able to run an integration test between typescript and javascript code, but for now, it's the best solution that allows you to be in control of the types and be sure that you are not breaking them during time.

### Acknowledgements

I would like to thank @brennanMKE and @dlmanning for giving the first stab at this and @fox1t for helping me review the code.

### TODO

We should do an extensive review to be sure that the entire API has been covered and that the definitions are correct.
@isaacs as you can see, I've added [`tsd`](https://github.com/SamVerschueren/tsd) as dev dependency, added a new script and updated the `test` script, let me know if everything is ok for you.


delvedor:
Hello @isaacs! Did you get a chance to review this pr? :)

delvedor (commented):
Hello @isaacs! Is there something more I can do here?

matteovivona:
Any news?

Ethan-Arrowood:
Would really love to use tap with my typescript projects. @isaacs I'm happy to do a thorough review of this PR if it helps getting it merged!

uwinkelvos:
Hey Guys!
It would be so nice if this was merged. Is the only holdup the `package-lock.json` merge conflict?

philp:
I'm going to guess @isaacs is maybe tied up with more important stuff right now ;) But agree, that as soon as this PR can be merged, it'll make writing tests much easier.

uwinkelvos:
@delvedor and at @isaacs: is it ok for you guys if I submit the typings to DT? We can always deprecate them as soon as this PR lands in master.

delvedor:
@uwinkelvos I don't like definitely typed, you are not in control of the types, and often they go out of date. But this is not my call.

types/test.d.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  export type DoesNotThrow = (fn?: (a: any) => any, message?: string, extra?: Options.Assert) => boolean
+
+  export type Equal = (found: any, wanted: any, message?: string, extra?: Options.Assert) => boolean
+
+  export type NotEqual = (found: any, notWanted: any, message?: string, extra?: Options.Assert) => boolean
+
+  export type Match = (found: any, pattern: any, message?: string, extra?: Options.Assert) => boolean
+
+  export type Type = (found: any, type: string | ((a: any) => any), message?: string, extra?: Options.Assert) => boolean
```
acdibble, line 184: The type should be `string` or a [constructor function](https://www.simonholywell.com/post/typescript-constructor-type.html):

```suggestion
  type Constructor<T extends {} = {}> = new (...args: any[]) => T
  export type Type = (found: any, type: string | Constructor, message?: string, extra?: Options.Assert) => boolean
```

types/test.d.ts
```diff
@@ -0,0 +1,203 @@
...
+    grep?: RegExp[],
+    only?: boolean,
+    runOnly?: boolean
+  }
+}
+
+export declare namespace Assertions {
+  export type Basic = (obj: any, message?: string, extra?: Options.Assert) => boolean
+
+  export type Throws = (fn?: (a: any) => any, expectedError?: Error, message?: string, extra?: Options.Assert) => boolean
```
acdibble, line 174: I think the following is more accurate to how `throws` can be used. I had to substitute `Error` for `Object` because of the example given in the documentation.
```suggestion
  type AnyFunction = (...args: any[]) => any
  export interface Throws {
    (fn: AnyFunction): boolean;
    (fn: AnyFunction, message: string): boolean;
    (fn: AnyFunction, expectedError: Object): boolean;
    (fn: AnyFunction, expectedError: Object, message: string): boolean;
    (fn: AnyFunction, expectedError: Object, extra: Options.Assert): boolean
    (fn: AnyFunction, message: string, extra: Options.Assert): boolean
    (fn: AnyFunction, expectedError: Object, message: string, extra: Options.Assert): boolean
  }
```

acdibble:
I created a [patch file](https://gist.github.com/acdibble/3ab3095e6c3d823f06692d9f25b79133) that can be used together with [patch-package](https://www.npmjs.com/package/patch-package) to apply the types in the node modules after install.

If anyone is looking for an alternative while this waits to be merged, this is pretty simple and useful.

types/test.d.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  export type DoesNotThrow = (fn?: (a: any) => any, message?: string, extra?: Options.Assert) => boolean
+
+  export type Equal = (found: any, wanted: any, message?: string, extra?: Options.Assert) => boolean
```
acdibble, line 178: Using generics will allow us to infer what properties `wanted` should have from `found`.
```suggestion
  export interface Equal {
    <T>(found: T, wanted: T): boolean;
    <T>(found: T[], wanted: T[]): boolean;
    <T>(found: T, wanted: T, message: string): boolean;
    <T>(found: T[], wanted: T[], message: string): boolean;
    <T>(found: T, wanted: T, extra: Options.Assert): boolean;
    <T>(found: T[], wanted: T[], extra: Options.Assert): boolean;
    <T>(found: T, wanted: T, message: string, extra: Options.Assert): boolean;
    <T>(found: T[], wanted: T[], message: string, extra: Options.Assert): boolean;
  }
```

ghost: Hmm, I'm not sure that use of generics is warranted here @acdibble. What's concerning is that if the actual JS runtime `typeof found` is, say, `"number"` but `typeof wanted` is, say, `"string"` then the generic type `T` would have to be `string | number` to pass static type checking (although guess the TS compiler would likely infer that if not explicitly given).

The purpose of comparing `found` against `wanted` for any kind of equality or inequality assertion is to verify runtime types and values and not compile-time static typings.

Actually, tbh, the first time I had a go at hacking together TypeScript typings for node-tap I dabbled with the generics route as you suggest above but soon abandoned generics for aforementioned reasons. Probably the most accurate TypeScript type for `found` is actually `unknown` as what the test found is always unknown until your test checks it!

The type annotation for `wanted` may as well be `unknown` also as having generic `<T>` arguably adds more technical debt than real value.

acdibble: I see your point. I was using specifically the `deepEqual` method when I was putting together these typings. I think the generic makes sense for that method, as it helps eliminate false assertions or even issues within the code itself.

```typescript
t.deepEqual<number[]>(myFunction(), [0,1,2,3,4]);
```

If `myFunction` doesn't return an array of numbers then you would get a compile-time error which could be helpful. I wasn't considering that this typing is used for several different methods, but I think there would therefore be value to more finely type the different assertion methods.

ghost: Thanks for engaging @acdibble. I see your point too & would love to see input from others to nail down the best typings in test scenarios!

uwinkelvos: I guess the typings do not necessarily need to reflect what makes sense, but what the untyped code allows. And in this case it would allow to compare apples with oranges. We know the outcome though... So I vote against the generics here. 

Btw. I started to prepare the typings for submission to DefinetlyTyped, including a test. Although I agree with @delvedor, that DT is not the best place for typings, it's still the second best option. Will share my branch here in the upcoming days.

types/test.d.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  export type NotEqual = (found: any, notWanted: any, message?: string, extra?: Options.Assert) => boolean
+
+  export type Match = (found: any, pattern: any, message?: string, extra?: Options.Assert) => boolean
+
```
nweldev, line 183:
```suggestion
export type MatchSnapshot = (found: any, message?: string, extra?: Options.Assert) => boolean
```

types/test.d.ts
```diff
@@ -0,0 +1,203 @@
...
+  unsimilar: Assertions.Match
+  notSimilar: Assertions.Match
+  unlike: Assertions.Match
+  isUnlike: Assertions.Match
+  notLike: Assertions.Match
+  isNotLike: Assertions.Match
+  doesNotHave: Assertions.Match
+  isNotSimilar: Assertions.Match
+  isDissimilar: Assertions.Match
+
```
nweldev, line 133:
```suggestion

matchSnapshot: Assertions.MatchSnapshot

```

nweldev (commented):
Snapshots are currently missing in test.d.ts.

Here is how I updated the patch for my project.

uwinkelvos:
Hey guys! I took delvedors typings, put in the documentation from node-tap.org and considered your remarks. This is the result:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43974
Please review and comment!

uwinkelvos:
Types are available on DT and already published: https://www.npmjs.com/package/@types/tap

jcao219:
Waiting for this

Having this in DT is fine for now, but we should definitely do a merge on the package-lock.json

jcao219:
DefinitelyTyped is a good place to store it right now, but we should add the types to the actual package a.s.a.p.

DefinitelyTyped does mean types can drift

However it also means the types are separated from the package itself, which is sometimes good

(For instance, you might be developing a NPM module that is type-only and you want to only `npm install` the types, not the code)

That's still not a good reason to prefer DT types to in-package types, so I made #681 

Thank you @delvedor for your hard work these are some top-notch types, can't wait to use them

types/tap.d.ts
```diff
@@ -0,0 +1,5 @@
+import { Tap } from './test'
+
+declare var tap: Tap;
+
+export = tap;
```
IlyaSemenov, line 5: I don't think that not having EOLs is in sync with the rest of node-tap source code?

zkldi:
Looks like DT's types are now out of date as of node-tap 15. Is anyone else currently working on updating these types, or would it be cool for me to start on it?

zkldi:
> 
> 
> Looks like DT's types are now out of date as of node-tap 15. Is anyone else currently working on updating these types, or would it be cool for me to start on it?

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52319

Should be good? Not too sure how I'd merge the changes from that .d.ts to @delvedor's PR, but more than willing to integrate this with the main repo.

PR #614, opened 2019-11-12
Exit with the yaml parse error on bad rc files

### LINKED ISSUE #613: .taprc parse failures should throw an error or at least warn
Creating an invalid `.taprc` does not result in any warning or error.  This can cause difficult to diagnose issues by causing tap to ignore settings which (from the user viewpoint) should be set.

isaacs:
Fix #613

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -123,7 +123,10 @@ const defaultFiles = options => new Promise((res, rej) => {
   })
 })
 
-const main = async options => {
+const main = options =>
+  mainAsync(options).catch(er => onError(er))
+
+const mainAsync = async options => {
   debug('main', options)
 
   if (require.main !== module)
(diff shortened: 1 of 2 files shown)


isaacs:
NB: a _missing_ rcfile is still fine, which does mean you could do `tap --rcfile=/this/does/not/exist`, which is clearly  a user error, but handling that while also not aborting on a missing default taprc file would be a tiny bit more involved.

bin/run.js
```diff
@@ -827,3 +829,10 @@ try {
     throw er
   }
 }
+
+try {
+  process.on('unhandledRejection', onError)
```
coreyfarrell: This is new, should `process.off` be added after the call to jack?

Otherwise this looks fine and I've verified it acts as I expect when `.taprc` is invalid.

isaacs: Turns out that removing the handler right after teh call to jack removes it before it can be relevant, because async function.

Updated branch with another approach that seems to work.

coreyfarrell: Retested with the updated branch, still resolves the issue for me.

PR #622, opened 2019-11-30
Pull initial TS_NODE_COMPILER_OPTIONS from test-env

IlyaSemenov:
Currently, this setup in package.json is half ignored:

```json
{
  "tap": {
    "test-env": [
      "TS_NODE_TRANSPILE_ONLY=1",
      "TS_NODE_COMPILER_OPTIONS={\"module\":\"commonjs\"}"
    ]
  }
}
```

the tests are run as:

```
  test: TAP
  env:
    TS_NODE_TRANSPILE_ONLY: "1"
    TS_NODE_COMPILER_OPTIONS: '{"jsx":"react"}'
```

After the fix, they are run as:

```
  test: TAP
  env:
    TS_NODE_TRANSPILE_ONLY: "1"
    TS_NODE_COMPILER_OPTIONS: '{"module":"commonjs","jsx":"react"}'
```

I understand that you will need regression tests etc. etc., consider this a heads up not the actual PR.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -652,7 +652,7 @@ const runAllFiles = (options, env, tap, processDB) => {
       if (options.ts && /\.tsx?$/.test(file)) {
         debug('ts file', file)
         const compilerOpts = JSON.stringify({
-          ...JSON.parse(process.env.TS_NODE_COMPILER_OPTIONS || '{}'),
+          ...JSON.parse(env.TS_NODE_COMPILER_OPTIONS || '{}'),
           jsx: 'react'
         })
         opt.env = {


isaacs:
This looks reasonable, but needs a test.  Do you have an example where this is relevant?

IlyaSemenov:
It's a bit complicated. My code base is partly shared between webpack build (where it needs `"module": "esnext"` due to runtime asynchronous imports) and 'normal' ts-node cli app (where it needs `"module": "commonjs"` to be able to even start). I set module: esnext in tsconfig (so that the IDE language server gets happy, too) but override it via environment vars when I run non-webpacked cli, including tap tests.

IlyaSemenov:
And while we are here, explicitly enabling jsx is way too opinionated, isn't it? Why don't you just let tsconfig.json define all the Typescript settings? If a user doesn't want and doesn't need JSX, why does tap insist on using JSX?

msuperina:
Hi I landed here after a bit of research and was wondering the same thing as @IlyaSemenov - why is tsconfig.json ignored ? Is there a CLI option unrelated to this code to pass the path of a tsconfig.json ? 
Many thanks for typescript support btw, I wasted 2 hours yesterday trying to run tape with typescript and automatic watch and eventually settled for tap where those are provided out of the box.

PR #625, opened 2019-12-05
Remove stdio-polyfill.js

### LINKED ISSUE #623: stdio-polyfill.js creates an infinite loop
Actual tapjs code at https://github.com/tapjs/node-tap/blob/master/lib/stdio-polyfill.js

Demo of the problem
```js
const {Writable} = require('stream')

delete process.stdout
process.stdout = new Writable({
  write (chunks, encoding, cb) {
    console.log('write:', chunks.toString('utf-8'))
    process.nextTick(cb)
  }
})

console.log('anything')
```

Calling `console.log` from the write function of `process.stdout` results in an infinite loop so running in this program results in 100% CPU utilization.  I noticed this while doing some digging into how `libtap` might allow an alternate output stream to be specified through `require('libtap/settings')`.

My current plan is to remove stdio-polyfill.js from libtap, replace with a check during module load (`settings.tapout` defaults to `process.stdout`):
```js
if (!settings.tapout || settings.tapout.destroyed || !settings.tapout.writable) {
  // Set exitCode in case process.stderr is also invalid
  process.exitCode = 1
  console.error('Output stream is invalid')
  process.exit(1)
}
```

coreyfarrell:
If this polyfill were run on node.js it will cause an infinite loop.

Fixes #623


lib/tap.js
```diff
@@ -20,11 +20,12 @@ const rootDomain = new Domain((er, type) => {
...
 })
 
-// We test that file separately, and removing stdout
-// causes some weird behavior in the Node test runner.
 /* istanbul ignore next */
 if (!process.stdout) {
-  require('./stdio-polyfill')()
+  // Set exitCode in case process.stderr is also invalid
+  process.exitCode = 1
+  console.error('Output stream is invalid')
```
coreyfarrell, line 27: @isaacs let me know if you want this message to be different or if you have any suggestion for how to test this.

isaacs: I guess it could be done with a child process, but honestly, it's such a weird scenario that this is probably fine.

isaacs (approved):
lgtm, but I'm using this PR to poke at a way to programmatically mark as merged, so please don't squash it in just yet :)

PR #627, opened 2019-12-09
Add node.js 13 to testing

coreyfarrell:
Normalize output changes to match node.js 12.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 
 node_js:
+  - "node"
   - 12
   - 10
   - 8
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: node_js
 
 node_js:
+  - 13
```
isaacs: Please add `node` to this list as well (or instead of 13) so that it's future-proofed.

isaacs:
LGTM 👍

.travis.yml
coreyfarrell: Done but when I get a chance I'm likely going to create a "tapjs/travis-ci" repository with shared travis configs similar to https://github.com/cfware/travis-ci (what I use for my repos).  This way testing config can be centrally managed for the whole tapjs org.

PR #631, opened 2019-12-13
Use libtap

coreyfarrell:
I believe the only potentially use-facing change is removal of
`TAP_DEV_SHORTSTACK` environmental variable.  Technically this has no
effect on end users as it was always the default.  `TAP_DEV_LONGSTACK`
still exists and allows users to strip tapjs and other npm "internals"
from stack traces.

Edit: This also bumps minimum node.js to v10.


lib/synonyms.js
```diff
@@ -73,7 +73,5 @@ module.exports = multiword({
     'notLike', 'isNotLike', 'doesNotHave', 'isNotSimilar', 'isDissimilar'
   ],
 
-  type: [
-    'isa', 'isA'
-  ]
+  type: ['isA']
```
coreyfarrell, line 76: Removed `isa` as the multiword function was generating it from `isA` so we ended up with duplicate of `isa`.

lib/tap.js
```diff
@@ -1,253 +1,34 @@
 'use strict'
-const Test = require('./test.js')
-const Stdin = require('./stdin.js')
-const Spawn = require('./spawn.js')
-const util = require('util')
-const objToYaml = require('./obj-to-yaml.js')
-const _didPipe = Symbol('_didPipe')
-const _unmagicPipe = Symbol('_unmagicPipe')
-const Domain = require('async-hook-domain')
+const settings = require('../settings.js')
```
coreyfarrell, line 2: Loading `../settings.js` before `libtap` is mandatory as it edits `libtap/settings` before the singleton is created by `libtap`.  For a user test which wants to edit settings this would happen as follows:
```js
const settings = require('tap/settings')
// Alter settings object
const t = require('tap')
```

If using `libtap` then the above code would use `libtap/settings` and `libtap`.  The following would be wrong:
```js
const settings = require('libtap/settings')
// Alter settings object
const t = require('tap')
```

In this case `require('tap')` would cause `tap/settings` to load for the first time, performing manipulations after the user changes.  This is an edge case as I think most users would not alter the settings.  Project-level alterations to settings can be done with a `--require` script through `--node-arg`:
```js
const settings = require('tap/settings')
// Alter settings object for all tests
```

settings.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
+
+const sourceMapSupport = require('source-map-support')
+const settings = require('libtap/settings');
+
+sourceMapSupport.install({environment:'node', hookRequire: true})
+
+if (+process.env.TAP_DEV_LONGSTACK !== 1) {
```
coreyfarrell, line 8: Scrutiny here might be useful.  Could probably add `append-transform` and `caching-transform` to ignoredPackages.

settings.js
```diff
@@ -0,0 +1,26 @@
...
+    'esm',
+    'nyc',
+    'import-jsx',
+    'function-loop'
+  )
+  settings.stackUtils.internals.push(
+    /at Generator\.next \(<anonymous>\)/iu
+  )
+} else {
+  settings.atTap = true
```
coreyfarrell, line 21: By default `libtap` will strip `at:` if it points into `libtap` but does not add `libtap` to the ignoredPackages.  Maybe this should be tweaked so we just check for `libtap` in the ignoredPackages instead of having a separate `atTap` setting?  Maybe `libtap` could be altered to strip any `at:` which matches something in `ignoredPackages` but that might be more difficult to implement and test.

tap-snapshots/test-run-save-file.js-TAP.test.js
```diff
@@ -96,6 +96,8 @@ not ok 2 - a/b/f1.js # {time}
         //f1.js
           require("{CWD}/").fail('a/b')
         --^
+      stack: |
```
coreyfarrell, line 99: I was not able to trace out why these stack's in this snapshot file are no longer completely stripped.

coreyfarrell (commented):
Definitely recommend selecting `Hide deleted files` if you view this in the github interface.  I've commented on a couple places that I feel are worth attention.

isaacs:
Landed on fea105cf30e19fff9ac2bddc1ac8d2c26f834f15, with a bit of squashing.

Highly considering just getting rid of the synonyms entirely, or maybe deprecating them somehow for deletion in v16, idk.  They're kind of a pita.

coreyfarrell:
> Highly considering just getting rid of the synonyms entirely, or maybe deprecating them somehow for deletion in v16, idk. They're kind of a pita.

@isaacs see #649 for this deprecation.  BTW not sure what's up with CI not running.

PR #635, opened 2019-12-20
Tapjs/node-tap/Promise plan

Hackdemon23:
Merge these branche can be right set it


isaacs:
Really not sure what this is doing, but at least, you'll have to rebase it onto main.

PR #642, opened 2020-01-24
add bowling to tap 100

tphummel:


tphummel:
this test failure doesn't appear to have anything to do with my markdown change. i don't have permissions to retry the build

PR #662, opened 2020-04-08
Add other operating systems to .travis.yml

christianbundy:
Problem: I've noticed some Windows-specific test failures that look
confusing, and it would be great to confirm that these failures aren't
coming from node-tap itself. After seeing a Windows-specific bug fix
(a9bd8f3) it seems like maybe if there were other Windows-specific bugs
we wouldn't know about them until someone brought them up.

Solution: Add macOS and Windows to `.travis.yml` to get automated
testing on those platforms as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ node_js:
 
 os:
   - linux
+  - osx
+  - windows
 
 cache:
   directories:


christianbundy:
Sweet, test failures! On macOS, these are failing:

- re-run a test when it changes
- re-run a test when covered file changes, root
- re-run a test when covered file changes, children
- filtered files

It looks like Windows is reporting 10 pending tests and 57 failures? The reporter wasn't working on Windows via Travis, so I had to download and edit the logs and pass it through a reporter manually. I wouldn't be surprised if I'm doing something Very Wrong. :innocent:

coreyfarrell:
tap is in the process of being split into a core test library and the test runner for v15.  After many attempts I got most tests working in the core library at https://github.com/tapjs/libtap/pull/15.

Most fixes were for tests but it did find one potential error in production code, namely that we split a stack trace message by `':'` to find the filename which would cause the win32 drive letter to be seen as the filename.  I suspect the fix is to instead split by `': '`.  I plan to submit that as a separate fix and backport to tap v14.

In this repository it might be easier to enable all operating system testing on the v15 branch as it has less tests than the master branch (many tests are moved to the libtap repository).  Any production bugs identified by Windows testing can be backported to tap v14.

jsumners:
I think this can be closed since #718 was merged.

isaacs:
Not using Travis any more, closing this.

PR #668, opened 2020-04-25
esm wrapper for tap

MylesBorins:
This is a sketch of a way to support native ESM for tap
which would be allowing for named exports when importing
tap 🎉.

Needs tests + docs, but wanted to get buy in for basic approach.

Also need to work out what interface is exposed in esm and how it is
exposed. Right now I'm exposing enough of a grab bag to be able
to refactor an esm test suite I have.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,11 +63,11 @@
   "repository": "https://github.com/tapjs/node-tap.git",
   "scripts": {
     "snap": "bash scripts/snap.sh",
-    "test": "node bin/run.js test -M coverage-map.js",
+    "test": "node bin/run.js -M coverage-map.js",
     "unit": "bash scripts/unit.sh",
     "posttest": "rm -rf cli-tests-*",
     "postunit": "npm run posttest",
-    "t": "node bin/run.js test -J -sfails.txt",
+    "t": "node bin/run.js -J -sfails.txt",
     "preversion": "npm test",
     "postversion": "npm publish",
     "postpublish": "bash postpublish.sh",
@@ -78,7 +78,7 @@
   },
   "tap": {
     "esm": false,
-    "test-ignore": "/((^|/)cli-tests-[0-9]+|\\.mjs$)/",
+    "test-regex": "^test\/.*\\.js$",
     "check-coverage": true
   },
   "nyc": {
--- a/tap-snapshots/test-run-dump-config.js-TAP.test.cjs
+++ b/tap-snapshots/test-run-dump-config.js-TAP.test.cjs
@@ -45,8 +45,8 @@ show-process-tree: false
 statements: 100
 test-arg: []
 test-env: []
-test-ignore: /((^|/)cli-tests-[0-9]+|\\.mjs$)/
-test-regex: ((\\/|^)(tests?|__tests?__)\\/.*|\\.(tests?|spec)|^\\/?tests?)\\.([mc]js|[jt]sx?)$
+test-ignore: $.
+test-regex: ^test/.*\\.js$
 timeout: 30
 ts: false
 version: false
@@ -96,8 +96,8 @@ show-process-tree: false
 statements: 100
 test-arg: []
 test-env: []
-test-ignore: /((^|/)cli-tests-[0-9]+|\\.mjs$)/
-test-regex: ((\\/|^)(tests?|__tests?__)\\/.*|\\.(tests?|spec)|^\\/?tests?)\\.([mc]js|[jt]sx?)$
+test-ignore: $.
+test-regex: ^test/.*\\.js$
 timeout: 30
 ts: false
 version: false
@@ -300,8 +300,8 @@ show-process-tree: false
 statements: 100
 test-arg: []
 test-env: []
-test-ignore: /((^|/)cli-tests-[0-9]+|\\.mjs$)/
-test-regex: ((\\/|^)(tests?|__tests?__)\\/.*|\\.(tests?|spec)|^\\/?tests?)\\.([mc]js|[jt]sx?)$
+test-ignore: $.
+test-regex: ^test/.*\\.js$
 timeout: 0
 ts: false
 version: false
@@ -364,8 +364,8 @@ test-arg:
   - xyz
   - abc
 test-env: []
-test-ignore: /((^|/)cli-tests-[0-9]+|\\.mjs$)/
-test-regex: ((\\/|^)(tests?|__tests?__)\\/.*|\\.(tests?|spec)|^\\/?tests?)\\.([mc]js|[jt]sx?)$
+test-ignore: $.
+test-regex: ^test/.*\\.js$
 timeout: 99
 ts: false
 version: false
@@ -415,8 +415,8 @@ show-process-tree: false
 statements: 100
 test-arg: []
 test-env: []
-test-ignore: /((^|/)cli-tests-[0-9]+|\\.mjs$)/
-test-regex: ((\\/|^)(tests?|__tests?__)\\/.*|\\.(tests?|spec)|^\\/?tests?)\\.([mc]js|[jt]sx?)$
+test-ignore: $.
+test-regex: ^test/.*\\.js$
 timeout: 30
 ts: false
 version: false


package.json
```diff
@@ -8,6 +8,13 @@
     "tap": "bin/run.js"
   },
   "main": "lib/tap.js",
+  "exports": {
+    ".": {
+      "import": "./lib/tap.mjs",
+      "default": "./lib/tap.js"
```
MylesBorins: This should be backwards compatible to most versions of Node.js. Any versions that don't understand package.exports will get main. All versions that supported exports support the default condition.

exporting `"./": "./"` maintains deep traversal of the module to keep this change `Semver-Minor` but you could start locking down the interface of the package if you wanted to in a future major

lib/tap.mjs
```diff
@@ -0,0 +1,22 @@
...
+const {
+  afterEach,
+  beforeEach,
+  Spawn,
+  Test,
+  test,
+  t
+} = tap;
+
+export default tap;
```
MylesBorins: Unsure if we want to export the default

coreyfarrell: Yes definitely.

coreyfarrell:
Conditional exports are implemented in `libtap`, could probably copy the `lib/tap.mjs` from that module.

Probably can use the testing added for that change as well, I used self reference imports to verify the package.json exports do as expected.

MylesBorins:
@coreyfarrell done re: copying from libtap.

What are your thoughts for testing / docs?

coreyfarrell:
@MylesBorins https://github.com/tapjs/libtap/blob/master/test/libtap.mjs is what I have for libtap (obviously the settings and versions tests do not apply here).  This verifies that ESM provides the expected named exports and that they match what is provided by CJS.  It uses self-reference in an attempt to prevent bugs caused by incorrect `package.json#exports`.

That test file is only executed in node.js 13.10.0+, doing this probably requires creating an `esm.js` to conditionally execute `test/tap.mjs` using `t.spawn` while adding `test/tap.mjs` to the `tap.test-ignore` regexp in package.json.

As for documentation I'm not sure what is needed, documentation is definitely an area of weakness for me.  We probably need to recommend setting `esm: false` as this could conflict (and seems to be causing issues in general).  The intent is that the next version of tap will not enable the `esm` user-space module by default.

One other comment the current `lib/tap.mjs` does not export any assertion aliases, see [asserts](https://node-tap.org/docs/api/asserts/) documentation.  So `const {assertNot} = require('tap')` works but `import {assertNot} from 'tap'` will not.  These aliases are being deprecated in the next semver-major of tap and since ESM named exports are a new feature I think should not have named exports for aliases in the first place; @isaacs please feel free to chime in on this.

MylesBorins:
@coreyfarrell got basic tests in place PTAL. There is a CJS wrapper that includes a fairly basic feature sniff and then shells out the ESM test if available. As of right now the tests won't run on 12, but will run on 13 and 14. If we were to add the `--experimental-modules` flag the tests would start working on 12.16.x

My guess is this isn't exactly how we would want to do this, especially since it doesn't seem to give us coverage... so I look forward to review on how to improve this.

test/test-esm.js
```diff
@@ -0,0 +1,28 @@
...
+try {
+  t = require('tap')
+} catch (e) {
+  t = require('../');
+  t.equals(e.code, 'MODULE_NOT_FOUND')
+  t.end();
+  return
+}
+
+const {spawn} = require('child_process')
```
coreyfarrell: I think we should use `t.spawn` instead of child_process.spawn.  This will cause tap output from `test.mjs` to be processed as a sub-test.

package.json
```diff
@@ -66,11 +73,11 @@
   "repository": "https://github.com/tapjs/node-tap.git",
   "scripts": {
     "snap": "bash scripts/snap.sh",
-    "test": "node bin/run.js test -M coverage-map.js",
+    "test": "node bin/run.js test/**/**.js test/*.js -M coverage-map.js",
```
coreyfarrell: I worry about using glob patterns on CLI.  Why not alter the `test-ignore` pattern in the `tap` section of package.json?

Even if we use these patterns these seem a bit wrong.  I think `test/**/*.js` should cover everything?  I think the patterns would need to be quoted if passed via CLI to prevent inappropriate processing by the shell, I'm not sure if the cross-platform way to quote them would be single-quotes `'` or double quotes `"`.

coreyfarrell (changes requested):
> @coreyfarrell got basic tests in place PTAL. There is a CJS wrapper that includes a fairly basic feature sniff and then shells out the ESM test if available. As of right now the tests won't run on 12, but will run on 13 and 14. If we were to add the `--experimental-modules` flag the tests would start working on 12.16.x

I think just testing on versions with unflagged modules is fine.  Also the tests require self-reference specifiers.

> My guess is this isn't exactly how we would want to do this, especially since it doesn't seem to give us coverage... so I look forward to review on how to improve this.

Coverage of lib/tap.mjs is tricky for tap v14.  This is because tap v14 uses nyc 14 but [@istanbuljs/esm-loader-hook](https://github.com/istanbuljs/esm-loader-hook) targets nyc 15.  In theory a simplified version of that loader hook could be vendored into this repository using babel-plugin-istanbul@5.  `lib/tap.mjs` doesn't have any actual code beyond the single `export const {...} = tap` so I'm not sure it's worth worrying about.

MylesBorins:
@coreyfarrell PTAL, finally got to updating this

coreyfarrell (changes requested):
Given `exports` no longer supports sub-path exports I think this change cannot be merged to tap v14 as it must be semver-major?

I'm thinking this needs to be re-targeted to the v15 branch.

MylesBorins:
@coreyfarrell subpath exports are still supported but deprecated on 15 (not 12 or 14). You can recreate the exact same behavior with `subpath patterns`, but that feature is only supported on the latest version of 12.x

```

{
  "exports": {
    "./*": "./*"
  }
}

TBH I still feel like this could land as a Semver-Minor in its current state but understand if you feel differently

**Comment by coreyfarrell** [comment]
@MylesBorins could we provide both `"./": "./"` and `"./*": "./*"` without causing spurious warnings in any version of node.js?  This way tap v14 could get this PR without breaking changes and tap v15 could be fully restricted exports (no subpath export or pattern).

**Comment by MylesBorins** [comment]
@coreyfarrell it works! If you define `"./*` first older versions of Node (that only support subpath folder) ignore the entry but match it before `"./"` avoiding the warning.

The current exports defined will now work on 12 / 14 / 15 with no warnings and also continue to support versions of Node that don't support exports through main 🎉

File: package.json [hunk 5]
```diff
@@ -8,6 +8,14 @@
     "tap": "bin/run.js"
   },
   "main": "lib/tap.js",
+  "exports": {
+    ".": {
+      "import": "./lib/tap.mjs",
+      "default": "./lib/tap.js"
+    },
```
ljharb:
```suggestion
    ".": [
      {
        "import": "./lib/tap.mjs",
        "default": "./lib/tap.js"
      },
      "./lib/tap.js"
    ],
```
this is required for node v13.0 and v13.1.

guybedford: Please don't encourage this pattern.

ljharb: I will continue to do so, and will be using it on all 300 of my packages as i gradually updated them to have “exports”, since we chose not to backport patches to those minors.

Since this package lacks an engines field, these node versions are supported, so not accepting this change makes this semver-major, even with the slash pattern fallback.

GeoffreyBooth: Node 13 is already past end-of-life.

ljharb: Which means node doesn’t support it anymore, but that doesn’t free packages from semver. Most of my packages support down to node 0.4, and will continue to do so for the foreseeable future.

package.json
```diff
@@ -81,7 +89,10 @@
   },
   "tap": {
     "esm": false,
-    "test-ignore": "/(^|/)cli-tests-[0-9]+/",
+    "test-ignore": [
+      "/(^|/)cli-tests-[0-9]+/",
+      "test/test.mjs"
+    ],
```
coreyfarrell: Unfortunately this is not an array (see test failure under node.js 8 / 10 where test/test.mjs is run).
```suggestion
    "test-ignore": "test-ignore": "/((^|/)cli-tests-[0-9]+|\\.mjs$)/",,
```

package.json
MylesBorins: @ljharb I don't agree with that assessment. While an engines field can be used to infer a support matrix I think it is a bit of a stretch to imply a lack of one implies support for all version. AFAIK there is no easy way in an engines field to say "this supports 10, 12, 14, 15, but not 11 and 13". 

Also fwiw the package has an engines field that says ">=8"

package.json
coreyfarrell: Sorry apparently I broke the JSON.  I'd love to blame github UI but that's probably not reasonable.

package.json
coreyfarrell: I also do not feel this is something which needs to be addressed.  Node.js 13.0 and 13.1 are broken, node.js 13.2.0+ is not.  If this broke all of 13.x I'd feel differently but the release family still works (just not the broken minor releases).

package.json
coreyfarrell: @MylesBorins I'm noticing now the exclude pattern ends up in some of the snapshots so they probably have to be regenerated.

package.json
ljharb: @mylesborins That’s all fine; you asked me to open this suggestion so you’d see what it would look like.

Thanks, I’d missed the engines field; the easiest way is `^10 || ^12 || ^14 || ^15`, ftr.

coreyfarrell:
@MylesBorins I've done some experimentation with this branch locally, it seems we found an issue where the `test-ignore` option is not honored if a path is given to the tap executable as the tap testsuite does (#708).

I pulled your branch and dropped the `test` argument from the npm `test` and `t` scripts, used `test-regex` instead of `test-ignore` to work around this issue.  If it's OK with you I'd like to push that to your branch to get the tests passing here.

MylesBorins:
Push away!

MylesBorins:
@coreyfarrell quick ping... what's needed to move this forward?

isaacs:
Landed on v15.  Thanks!

PR #674, opened 2020-05-20
Removed redundant RegExp escaping

vanodevium:


isaacs:
I'm going to punt on any and all stylistic changes until we get a fully configured linter that can run with `--fix`.  It'll catch these.

PR #680, opened 2020-06-27
upgrade foreground-child

FauxFaux:
Drops support for node <8, other minor fixes.

https://github.com/tapjs/foreground-child/blob/master/CHANGELOG.md#v200

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -685,21 +685,32 @@
       }
     },
     "cross-spawn": {
-      "version": "4.0.2",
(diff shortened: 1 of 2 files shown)


coreyfarrell:
Although tap can upgrade foreground-child as a non-breaking change nyc 14 uses foreground-child 1.x.  nyc 15 upgrades foreground-child but I think an nyc upgrade here would be considered breaking.  I'm not sure this update makes sense for tap v14 as foreground-child 1.x will still be installed and used.

Note the v15 branch (WIP for the next version of tap) does this upgrade already.

Issue #682, opened 2020-06-28
ERR_UNKNOWN_FILE_EXTENSION using node >= 12.17


simone-sanfratello:
Hi,

try running tap v. 14.10.7 with node v. 12.18.1 and node v. 14.4.0, I get the following error

![err](https://user-images.githubusercontent.com/5823031/85950857-98836100-b95f-11ea-9f19-5b95eacbf3ac.png)

Everything works using node v. 12.16.3, I guess because in 12.17 they introduce "ECMAScript Modules - --experimental-modules flag removal"

simone-sanfratello:
submitted PR https://github.com/tapjs/node-tap/pull/683

PR #683, opened 2020-06-28
chore: update supported node version

simone-sanfratello:
follow https://github.com/tapjs/node-tap/issues/682

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "main": "lib/tap.js",
   "engines": {
-    "node": ">=8"
+    "node": ">= 8 < 12.17"
   },
   "dependencies": {
     "@types/react": "^16.9.16",


coreyfarrell:
We're dealing with a conditional error, I just created a minimal test:
```
mkdir issue-682
cd issue-682
npm init -y
npm i tap

cat > index.js << EOF
'use strict';
module.exports = 'index.js';
EOF

cat > test.js << EOF
'use strict';
const t = require('tap');
const mod = require('./index.js');

t.equal(mod, 'index.js');
EOF
```

Running `npx tap` ran the test without error (on node.js 12.18.0 and 14.4.0).  I'm closing this PR as it's premature and unlikely to be accepted.  Let's move discussion to the issue and try to figure out the conditions which create a problem.  We'll try to fix the underlying issue or document the caveat.  Documenting tap as incompatible with Node.js >=12.17 would only be considered if it's unconditionally broken and Node.js refuses to fix.

PR #698, opened 2020-10-16
Add mock API

ruyadorno:
This brings into **tap** the standard mocking system we have been using across the ecosystem of packages from the npm cli which consists into hijacking the `require` calls from a given module and defining whatever mock we want via something as simple as a key/value object.

It's a very conscious decision to make it a very opinionated API, as stated in https://github.com/tapjs/node-tap#tutti-i-gusti-sono-gusti - focusing only on the pattern that have been the standard way we handle mocks.

It builds on the initial draft work from @nlf (ref: https://gist.github.com/nlf/52ca6adab49e5b3939ba37c7f0fc51c6) and initial brainstorming of such an API with @mikemimik - thanks ❤️ 

### Example

```js
t.test('testing something, t => {
  const myModule = t.mock('../my-module.js', {
    fs: { readFileSync: () => 'foo' }
  })

  t.equal(myModule.bar(), 'foo', 'should receive expected content')
})
```

### DIFF
--- a/lib/mock.js
+++ b/lib/mock.js
@@ -1,9 +1,13 @@
 const Module = require('module')
 
 class Mock {
-  constructor(filename, mocks) {
+  constructor(parentFilename, filename, mocks) {
     this.filename = filename
-    this.mocks = mocks
+    this.mocks = new Map()
+
+    if (!parentFilename || typeof parentFilename !== 'string') {
+      throw new TypeError('A parentFilename is required to resolve Mocks paths')
+    }
 
     if (!filename || typeof filename !== 'string') {
       throw new TypeError('t.mock() first argument should be a string')
@@ -14,10 +18,16 @@ class Mock {
 are the same used in ${filename} require calls`)
     }
 
-    const callerTestRef = module.parent.parent.parent
+    const callerTestRef = Module._cache[parentFilename]
     const filePath = Module._resolveFilename(filename, callerTestRef)
     if (!Module._cache[filePath]) {
       Module._load(filename, callerTestRef, false)
+    }
+
+    // populate mocks Map from resolved filenames
+    for (const key of Object.keys(mocks)) {
+      const mockFilePath = Module._resolveFilename(key, callerTestRef)
+      this.mocks.set(mockFilePath, mocks[key])
     }
 
     this.original = Module._cache[filePath]
@@ -28,11 +38,14 @@ are the same used in ${filename} require calls`)
       this.original
     )
 
-    mod.require = function requireMock(id) {
-      if (Object.prototype.hasOwnProperty.call(mocks, id))
-        return mocks[id]
+    const mock = this
+    mod.require = function tapRequireMock(id) {
+      const requiredFilePath = Module._resolveFilename(id, mock.original)
+      const res = mock.mocks.get(requiredFilePath)
+      if (res)
+        return res
 
-      return originalRequire.call(this, id)
+      return mock.original.require.call(this, id)
     }
 
     mod.loaded = false
@@ -40,8 +53,8 @@ are the same used in ${filename} require calls`)
     this.module = mod.exports
   }
 
-  static get(filename, mocks) {
-    const mock = new Mock(filename, mocks)
+  static get(parentFilename, filename, mocks) {
+    const mock = new Mock(parentFilename, filename, mocks)
     return mock.module
   }
 }
--- a/lib/mock.js
+++ b/lib/mock.js
@@ -1,7 +1,7 @@
 const Module = require('module')
 
 class Mock {
-  constructor(parentFilename, filename, mocks) {
+  constructor(parentFilename, filename, mocks = {}) {
     this.filename = filename
     this.mocks = new Map()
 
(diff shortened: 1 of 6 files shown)


lib/mock.js
```diff
@@ -0,0 +1,48 @@
...
+    if (!filename || typeof filename !== 'string') {
+      throw new TypeError('filename should be a string')
+    }
+
+    if (!mocks || typeof mocks !== 'object') {
+      throw new TypeError(`mocks should be a a key/value object in which keys
+are the same used in ${filename} require calls`)
+    }
+
+    const filePath = Module._resolveFilename(filename, module.parent)
```
isaacs: Won't `module.parent` be something in tap?  Seems like it should be resolved from the test file, no?

isaacs: Maybe what we need to do is climb up `module.parent.parent....` until we find the main `tap` module, and use _its_ parent?

isaacs:
```js
let p = module.parent
while (p && p is not the tap root module figure that out somehow) {
  p = p.parent
}
const filepath = Module._resolveFilename(filename, p.parent) // <-- this is whatever require()'d tap
```

isaacs:
```js
// test/lib/my-test.js
const t = require('tap')
const myLIb = t.mock('../../lib/my-thing.js', {
  fs: myFsMock,
  // mock for require('../../lib/my-other-thing.js')
  '../../lib/my-other-thing.js': { ... }, // path from this test module? require-inject style
  './my-other-thing.js': { ... }, // path from the target module?
})
```

isaacs: Actually, maybe we could just capture a stack trace, and use the path to whichever file is calling the `t.mock()` method that way?  Then it wouldn't matter who loaded `tap` first.

isaacs (changes requested):
I like this!  Only comment is that the module and the mock keys should be modules relative to the file calling `t.mock()`, not `tap` itself.  (Basically, should work the same as `require-inject` in that way, since while it does have a bit of a learning curve, it's pretty easy to grok and harder to get wrong, once it is learned.)

coreyfarrell:
I'm a bit concerned with having something in tap that is dependent on tested code being CJS.  I think a warning that this will not work on ESM modules needs to be added to the documentation.

Another question, what is the benefit of `t.mock` over packaging this as an auxiliary module where users would `require('tap-mock')`?  Even if tap has `t.mock = require('tap-mock')` my bias tends towards isolating stuff into smaller modules as I find they get tested better.

Not intending to be a blocker on this but I wanted to raise these points.

ruyadorno:
> I'm a bit concerned with having something in tap that is dependent on tested code being CJS. I think a warning that this will not work on ESM modules needs to be added to the documentation.

sure! I do want to have esm support too and would be willing to follow up with the work to get it there but I guess esm has a bunch of different discussions such as supporting only dynamic imports, etc and I would love to get `t.mock()` cjs usable before jumping down that other rabbit hole.

> Another question, what is the benefit of t.mock over packaging this as an auxiliary module where users would require('tap-mock')? Even if tap has t.mock = require('tap-mock') my bias tends towards isolating stuff into smaller modules as I find they get tested better.

I guess this is the very opinionated and personal-preference aspect of the developer experience involved around using the test runner, I for one much prefer the DX of having these essential built-in features right there in `t` and a very good comparison I can think of is the addition of the Fixtures API (`t.testdir`, `t.fixture`, etc over `require('tacks')` and similars) - it has changed SO MUCH for the better my personal usage that it was the turning point for `tap` to become my default test runner for every new project - all that to say that my contribution to add `t.mock()` is also in the hope of steering the project a little bit more in favor of this kind of out-of-the-box complete test runner experience, that to a large extent `tap` already has by providing coverage out of the box, fixtures, snapshots, etc

Hope that makes sense to you @coreyfarrell and thanks, I appreciate your feedback 😄

lib/mock.js
ruyadorno: I can't really see how capturing a stack trace could be easier/simpler than looping through parents and checking for who loaded `tap` similar to the example you provided in the 3rd comment 😅

isaacs:
Here's how to get the filename of the file that's calling `t.mock()` so that all the paths can be resolved from there:

```js
// in lib/test.js
class Test {
  // ....
  mock (module, mocks) {
    const {file} = stack.at(Test.prototype.mock)
    const resolved = path.resolve(file)
    // console.error(resolved)
    // then resolve module paths from there, pass to the Mock object, etc.
  }
  // ....
}
```

EDIT: updated because I noticed that stack-utils has a helper method that does all of this, in a slightly more efficient way, setting the stackTraceLimit to 1 since we only need a single frame.

ruyadorno:
hey @isaacs I took the time to implement the suggested changes and clean things up over the weekend - also added many variations of the scenarios we chatted about to make sure module resolution works as intended in `test/test.js` 😊

lib/mock.js
isaacs: Example:

```js
// ./foo.js
module.exports = require('./bar.js')
```

```js
// ./bar.js
module.exports = 'not baz'
```

```js
// test/common/shared-test.js
const t = require('tap')
module.exports = () => t.mock('../../foo.js', { '../../bar.js': 'baz' })
```

```js
// test/first.js
const t = require('tap') // this is the module that loaded tap
const shared = require('./common/shared-test.js') // not this one
const mockFoo = shared()
t.equal(mockFoo, 'baz')
```

```js
// second-test.js
const shared = require('./test/common/shared-test.js') // this is the one that loads tap
const t = require('tap') // not this one
const mockFoo = shared()
t.equal(mockFoo, 'baz')
```

In `test/first.js`, the "module that loaded tap" is `test/first.js`.  In `second-test.js`, the "module that loaded tap" is `test/shared/common.js`.

So, really, you _have_ to walk up the stack and find the thing that's actually calling `t.mock()`.

Also, calling `stack.at(t.mock)` is arguably easier, and does not depend on module system internals.

isaacs: Oh, lol, gmail decided to bubble this to the top of my inbox and I thought it was a new comment.  Sorry to rehash something from last month, but I guess it's good to have for posterity anyway 😅

ruyadorno:
[update] I'm in the process of replacing all mock usage within the https://github.com/npm/cli codebase with this `t.mock` implementation and during that process I found some very specific edge cases which I'm currently tackling - I'll make sure to add a ref between the PR with the rewrote mocks for **npm** and this one once that's ready to go. 😊

ruyadorno:
Here's the PR that updates the npm cli to use `t.mock` instead: https://github.com/npm/cli/pull/2370

jordanbtucker:
Is it possible to mock the built-in `fetch` with this mock API?

PR #717, opened 2021-02-12
Add detection for .yaml and .yml

jsumners:
This resolves #706.

I based my tests on `test/run/bad-rcfile.js` because I didn't see any other tests that cover parsing of rc files that I could learn from.

I am not clear how to run the tests and generate an HTML coverage report so that I can see what branch(es) are not being covered. When I run `npm run unit` I can see that some branch coverage is missing, but not what it is.

PR #728, opened 2021-03-25
Tap version 15

### LINKED ISSUE #603: Todo list for tap v15
## node-tap changes
- [x] breaking: Remove `--esm` (#651)
- [x] breaking: Upgrade to nyc 15 (#650, tapjs/libtap#9)
- [x] breaking: default `ts: false` and `jsx: false` (#651)
- [x] breaking: default `check-coverage: true` (everyone joins the `--100` club!)
- [x] refactor: Split this lib into two modules: `libtap` (all the library bits that tests actually use) and `tap` (the runner, watcher, reporter, repl, nyc logic, etc.)
- [x] remove `source-map-support` from async-hook-domain (unnecessary, and the wrong place for this)
- [x] make the stack-utils API nicer
- [x] breaking: Drop support for node <= 8
- [x] breaking: change snapshot file extensions to `.cjs`
- [x] breaking: deprecate aliases (#649)
- [x] breaking: add conditional exports to support named ESM exports, forbid outside resolve of internal-only `tap` sources from other modules

## libtap changes
- [x] breaking: Separate `t.has` from `t.match`.  In general, make all `tcompare` methods map directly to a test assertion of the same name.
- [x] breaking: Upgrade to tcompare 4.x, DRY all diffing logic into that one location.  (No need to handle strings separately, etc.)
- [x] add inheritable `t.compareOptions` object to pass options to tcompare
- [x] refactor: Split `lib/test.js` into parts.  It's way too long.
- [x] put `t.testdir()` fixtures in a folder that is excluded by the default `test-regex` conf.  (Like, `./fixtures/${test name}` or something.)
- [x] breaking: add conditional exports to support named ESM exports, forbid outside resolve of internal-only `libtap` sources from other modules (tapjs/libtap#10)

isaacs:
Fix: https://github.com/tapjs/node-tap/issues/603

## 15.0 - 2021-03-30

This is a major refactor of much of tap's internals, and a lot of new
features.

### BREAKING CHANGES

* Drop the use of the `@std/esm` module, in favor of native ES Modules.
* Drop the inclusion of `typescript` by default.  (Typescript still
  supported, but requires that you install it yourself.)
* `.jsx` files only run automatically when `--jsx` config is explicitly
  enabled.
* `--check-coverage` on by default.
* Drop support for node `<10`.
* Separate `t.has` from `t.match`, so these are distinct.
* Deprecate aliases.
* Do not report on test points filtered with `only` or `grep` options.
* Resolve `t.test()` promise to the child test results, rather than the
  parent test.
* Remove `callback` argument from `t.beforeEach` and `t.afterEach`. Return
  a promise if you wish these methods to be async.

### NEW FEATURES and BUG FIXES

* Restructure snapshot output folder, and change file extensions to `.cjs`.
* Add `t.compareOptions` object to pass options to all the methods that use
  `tcompare` (ie, `t.has`, `t.match`, `t.same`, etc.)
* Improved diffing and comparison output for long strings and buffers.
* Add `t.before` method.
* Add `t.mock()` API for mocking calls to `require()` in modules being
  tested.
* Inherit the `t.saveFixture` boolean option.
* Create fixtures symbolic links as junctions if pointing at directories.
* Set both `FORCE_COLOR` and `NO_COLOR` environment variables
  appropriately.
* Pull initial `TS_NODE_COMPILER_OPTIONS` from test environment.
* Run fixture cleanup aysnchronously on `t.teardown()` to minimize Windows
  folder locking issues.

### DEPENDENCIES and REFACTORING

* Extract most of the internal functionality to
  [`libtap`](https://npm.im/libtap).
* Update `nyc` to version 15.
* Conditional exports to limit diving into tap's internals except via
  supported APIs.


isaacs:
This PR doesn't actually merge cleanly, it's mostly just here for folks to have a thing to comment on and see the changelog.  Maybe this is as good a time as any to switch the default branch from `master` to `main`, and rename the current `master` to `v14`?

jsumners:
Would it be possible to get #717 included?

gr2m:
I just upgraded to v15 after one of my dependencies changed to ES Modules:
https://github.com/wip/app/pull/323

Upgrade went smooth! The change from 

```js
const emojiToName = require("gemoji/emoji-to-name");
```

to 

```js
const { emojiToName } = await import("gemoji");
```

caused a weird error in tap 14, likely due to the usage of `@std/esm`. It went away after I upgraded to v15.

The deprecation messages were very helpful, great upgrade experience overall.

Plus, it feels snappy! Great work!

coreyfarrell:
> Plus, it feels snappy! Great work!

It is faster than v14.  This is partly due to improvements in nyc 15 compared to 14, but also due to removal of the `esm` userspace transpiler.

coreyfarrell (changes requested):
In the release notes I think enabling `--check-coverage` should be explained that tap will exit with error if coverage of statements/lines/branches/functions is less than 100%?  This might reduce the number of issues we see due to this change.

Not strictly needed for the release but I think we can/should fix the tests to run in node 10.  libtap run the `.mjs` tests conditionally by node version.  I'm not suggesting this conditional run logic should be part of what users get from `npm install tap` but tap should be testable on node 10.  No strong objection if node 10 has to be done with like `npm run test-legacy` or something just think it's important for CI to cover it.

jsumners:
> I think we can/should fix the tests to run in node 10.

I don't think so. Node 10 is EOL on 2021-04-30. I think hobbling `tap@15` with Node 10 support because it releases about a month before 10's EOL is not very useful.

isaacs:
> In the release notes I think enabling --check-coverage should be explained that tap will exit with error if coverage of statements/lines/branches/functions is less than 100%? This might reduce the number of issues we see due to this change.
> 
> Not strictly needed for the release but I think we can/should fix the tests to run in node 10. libtap run the .mjs tests conditionally by node version. I'm not suggesting this conditional run logic should be part of what users get from npm install tap but tap should be testable on node 10. No strong objection if node 10 has to be done with like npm run test-legacy or something just think it's important for CI to cover it.

Yeah, I agree.  I'd also like to see if it's possible to work further back on the 10.x series, or if not, we need to specify a minimum node v10 that we support.  The tricky thing about the mjs test is that it fails parsing on node 10, so we can't even just put a `skip` in there very easily.

coreyfarrell:
I'm not sure dropping node 10 early gets us much.  It's only useful if tap can pull node >=12 versions of dependencies without being semver-major for tap.  Specifically I'm thinking about nyc 16 which will use babel 8 and have more breaking changes than just the version requirements.  Until that's ready we can't know if tap will need a new semver-major to get those updated versions.

mcollina:
@coreyfarrell @isaacs could we remove request from the dependency chain?

isaacs:
Right now request is coming in via coveralls.  So either we accept it, drop coveralls support, or coveralls moves off request.

coreyfarrell:
At least one fork that moves off request, coveralls-next.  I don't know anything more about the fork, just that it exists.  Coveralls module does not seem to be actively maintained.

jsumners:
FWIW, I am using this in a private repo to get a coverage report in PRs:

```yaml
  check_coverage:
    name: Check Coverage
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-node@v2
        with:
          node-version: ${{ matrix.node }}
          registry-url: "https://registry.npmjs.org"
      - uses: actions/cache@v2
        with:
          path: ~/.npm
          key: ${{ runner.os }}-node-${{ hashFiles('**/package.json') }}
          restore-keys: |
            ${{ runner.os }}-node-
      - name: Install Packages
        run: npm install
        env:
          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
      - name: Generate lcov
        run: npm run test:lcov
      - uses: romeovs/lcov-reporter-action@v0.2.16
        with:
          github-token: ${{ secrets.GITHUB_TOKEN }}
          lcov-file: ./coverage/lcov.info
```

PR #730, opened 2021-03-31
test on all operating systems again

isaacs:


isaacs:
this will have to be revisited, unfortunately.

PR #732, opened 2021-04-01
Include tap.mjs in the released bundle

mcollina:
The `lib/tap.mjs` file is not present in the bundle, making tap unloadable in native esm.


isaacs:
Already landed based on twitter ping, sorry, didn't see this PR.

mcollina:
No worries, shipping fast was better!

PR #734, opened 2021-04-01
Fix deprecation warning for destructured aliases

jsumners:
Not sure this is even possible, but giving it a try.


jsumners:
Never mind. I was not seeing deprecation warnings in some Pino tests. Turns out, the subset _not_ giving the warnings are actually using `tape` since they target browser environments.

Issue #743, opened 2021-04-15
Cannot find module 'color-support'


jsumners:
`tap@15.0.2` error out with the error in this issue title.

**package.json**:
```json
{
  "dependencies": {
    "tap": "^15.0.2"
  }
}
```

**test.js**:
```js
'use strict'

const tap = require('tap')

tap.test('test', async t => {
  t.equal(1, 1)
})
```

Add those files, run `npm install`, and then `tap test.js` to get:

```
❯ tap test.js
Error: Cannot find module 'color-support'
Require stack:
- /private/tmp/15/tap-test/node_modules/tap/bin/jack.js
- /private/tmp/15/tap-test/node_modules/tap/bin/run.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)
    at Function.Module._load (internal/modules/cjs/loader.js:725:27)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/private/tmp/15/tap-test/node_modules/tap/bin/jack.js:4:22)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/private/tmp/15/tap-test/node_modules/tap/bin/jack.js',
    '/private/tmp/15/tap-test/node_modules/tap/bin/run.js'
  ]
}
```

```
❯ npm --version
6.14.11
```

jsumners:
Actually, it looks like specifying `"tap": "15.0.0"` will also result in the error. So it must be some transient dependency introducing the error.

jsumners:
No. I think it's a core problem. It looks like the dependency was removed in this commit -- https://github.com/tapjs/node-tap/commit/61e0e60a25f6b09d2e0152d4ea6f52b8e8382af4

But `jack.js` in the current branch requires it:

https://github.com/tapjs/node-tap/blob/23c7821213b44e2f85579ea5a244cd4c6ffe35df/bin/jack.js#L4

PR #745, opened 2021-04-19
Remove unused import

jsumners:
This is to solve issue #743. The issue actually only happens when a project has an `.npmrc` with `legacy-bundling=true` set. This causes `bin/jack.js` to not find the dependency because it hasn't been flattened out to the top-level of `node_modules`. As far as I know, it's effectively impossible to write a test for this case. However, the dependency is completely unused in the `bin/jack.js` code, so the fix is simple: remove the import.

### DIFF
--- a/bin/jack.js
+++ b/bin/jack.js
@@ -1,7 +1,6 @@
 const { jack, num, opt, list, flag, env } = require('jackspeak')
 
 const {cpus} = require('os')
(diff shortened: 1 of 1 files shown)


jsumners:
@isaacs or @coreyfarrell what do you think?

isaacs:
Thanks!

PR #748, opened 2021-05-03
Remove unused variables / add linting

jsumners:
In my last few PRs I have noticed that there are many unused variables across the code. This _can_ lead to issues as we saw in #745. This PR adds eslint to catch these sorts of errors. The eslint configuration is based on the standard "eslint:recommended" rule set. I have disabled:

+ `no-prototype-builtins`
+ `no-empty`
+ `no-regex-spaces`
+ `no-useless-escape`

I have disabled these so that this PR has _minimal_ impact on the code. I recommend someone who knows both the code and the regular expressions to re-enable the `no-regex-spaces` and `no-useless-escape` rules and fixing the code accordingly.

### DIFF
--- a/bin/jack.js
+++ b/bin/jack.js
@@ -6,7 +6,6 @@ const reporters = [...new Set([
   ...(require('tap-mocha-reporter').types),
   ...(require('treport/types')),
 ])]
-const fs = require('fs')
 // nyc bundles its deps, pull reporters out of it
 const nycReporters = [
   'clover',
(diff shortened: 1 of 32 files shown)


bin/jack.js
```diff
@@ -692,6 +691,7 @@ Much more documentation available at: https://www.node-tap.org/
                   appended to the filenames.`,
   }),
 
+  /* eslint-disable-next-line */
```
jsumners, line 695: I'm not sure what to do here. There's a lot of code in here and it's hard to see it all. So I just disable the warning. However, if it is accurate, then I think this key will take precedence over the first definition.

jsumners: To be clear: this comment directive is disabling a "duplicate key" warning.

jsumners:
ping: @isaacs @coreyfarrell

PR #757, opened 2021-05-31
docs: don't claim ts-node is bundled

midgleyc:
A docs-only change. `ts-node` isn't bundled any more, so don't claim that it is.


isaacs:
Ah, looks like this already got fixed.

PR #759, opened 2021-06-02
Create and include types.d.ts

zkldi:
Past PRs for this include #607.

**Motivation**
- `tap` is batteries included and including types reduces amount of places to look for node-tap stuff
- there's currently only me maintaining the types on DT which means every change requires bothering DT maintainers
- means any changes to tap can be kept in sync with its types easier

**Why not #607?**
node-tap got bumped from 14 to 15 recently and those types are mostly out of date (especially with assertion synonyms). I also modified parts of the types for better IDE integration and to fix some issues (such as `t.throws/t.match/t.hasStrict/t.rejects` wanting incorrect types)

**PR Outline**

- Creates `types/types.d.ts`
- Excludes it from the gitignore
- Adds `types: "types/types.d.ts"` to the package.json

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 
 !bin/
 !lib/
+!types/
 !settings.js
 !docs/
 /docs/public
--- a/package.json
+++ b/package.json
@@ -120,5 +120,6 @@
   ],
   "funding": {
     "url": "https://github.com/sponsors/isaacs"
-  }
+  },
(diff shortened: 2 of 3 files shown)


zkldi:
Actually, going to revoke this PR as mentioned in the other ones TS tests aren't optional and i'm not familiar enough to set this up.

if anyone comes across this thread, though, I am looking for a second maintainer for @types/tap.

PR #817, opened 2022-03-21
feat(cli): remove validation for --coverage-report

ahmadnassri:
this allows for custom nyc coverage report packages to be used.

reference: https://github.com/istanbuljs/nyc#common-configuration-options

> `reporter`: May be set to a built-in coverage reporter or an npm package (dev)dependency

### DIFF
--- a/bin/jack.js
+++ b/bin/jack.js
@@ -376,7 +376,6 @@ Much more documentation available at: https://www.node-tap.org/
 
   'coverage-report': list({
     hint: 'type',
-    valid: nycReporters,
     description: `Output coverage information using the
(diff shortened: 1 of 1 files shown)


isaacs:
Oh, I didn't realize it could be a dep name! It's probably still worth showing the built-in ones, so you're not surprised that it doesn't load the `lcov` module or something? Removing the validation is correct though, thanks.

ahmadnassri:
@isaacs good call, I edited the change to remove the validation only, and slightly modify the default message.

isaacs:
Published on 16.2, thanks!

PR #844, opened 2022-11-16
update workflows

Uzlopak:
dropping node 12 testing, as it is EOL
update actions to fix deprecation notices in workflow runs
adding node 19 testing


isaacs:
v16 does still need to be run on node 12.  Even though it's EOL, tap 16 supports it.

Uzlopak:
I added back 12

Uzlopak:
Closing due to inactivity

PR #845, opened 2022-11-16
move bundled dependencies to devDependencies

Uzlopak:
I assume that the current windows issues are a result of the bundling of the dependencies. So maybe to resolve this issue, we need to move the dependencies which are anyway bundled into devDepencencies.

I dont know how to test this assumption

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,14 +16,11 @@
       ],
       "license": "ISC",
       "dependencies": {
-        "@isaacs/import-jsx": "^4.0.1",
-        "@types/react": "^17",
         "chokidar": "^3.3.0",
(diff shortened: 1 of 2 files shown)


Uzlopak:
@isaacs 
@climba03003

isaacs:
Fixed in a different way.  Next major will likely require node 16 and npm 7+ anyway, so that's likely the last time it'll happen.  I have a test env setup to reproduce it now, if I ever do need to do another v16 release for some reason.

Uzlopak:
@isaacs 
Its more about not downloading again the react stuff if we already bundled it with the package. But i am excited to see your solution.

PR #863, opened 2023-04-23
Upgrade import-jsx

breck7:
I was hitting a `digital envelope routines::unsupported` error in `Object.createHash` which was being called from `import-jsx`.

I noticed @isaacs had forked that package a while back and switched it to md-5 from md-4.

It seems the package author has since made those updates as well as a few more fixes.

This switches back to the original package. It does fix the OpenSSL error, *however*, after I made the change I realized I am not sure if previously I was running an older version of Tap. So this may have already been fixed.

Nonetheless, I am submitting it anyway, as it seems like it is less code for you to maintain (no need for your own fork of import-tsx now).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,13 +16,13 @@
       ],
       "license": "ISC",
       "dependencies": {
-        "@isaacs/import-jsx": "^4.0.1",
         "@types/react": "^17.0.52",
         "chokidar": "^3.3.0",
         "findit": "^2.0.0",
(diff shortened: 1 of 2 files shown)


isaacs:
Sweet, thanks. I'll take a look soon and make sure there's nothing else that might require the fork to stick around. Iirc, it was just the hashing algo, so probably fine.

isaacs:
No longer relevant in next version.

PR #910, opened 2023-09-22
Isaacs/node test interop

### LINKED ISSUE #888: Interop with node.js test runner
Running tests with the [Node.js test runner](https://nodejs.org/api/test.html#running-tests-from-the-command-line) `node --test` seems to simply run all tap files.

In particular `node --test --test-name-pattern="no test matches this"` simply runs all files as simple scripts, without any filtering on the pattern.

To what extent is the Node.js test runner supported (is this planned? Does a plugin exist?)

isaacs:
Fix: #888


isaacs:
CI is actually passing, I just canceled the `push` action by mistake. But the `pull_request` action is identical.

PR #941, opened 2023-10-05
Comb through every combo of plan, promise, end()

### LINKED ISSUE #937: t.plan plus t.resolves (or t.rejects) causes "test assertion after Promise resolution" error
With tap >= 18.0.0 we have issues in our repositories which can be boiled down to these minimal cases:

```
import tap from 'tap'
tap.test('fails resolve', async t => {
  t.plan(1)
  await t.resolves(new Promise(r => setTimeout(r, 1000)))
})
```

```
> fails resolve > test assertion after Promise resolution
```

```
import tap from 'tap'
tap.test('fails reject', async t => {
  t.plan(1)  
  await t.rejects(Promise.reject(() => new Error('should reject')))
})
```

```
> fails reject > test assertion after Promise resolution
```

With tap 16.3.8, both tests above pass successfully.

isaacs:
When a test function returns a promise, and there's a promise-emitting assertion like t.resolves() or t.rejects() that triggers a `t.waitOn` call to occupy the test, it was failing to recognize that it's awaiting an implicit end, so when the occupying Waiter resolves, it would just stop and let the process exit, triggering a "test not finished" error in some cases.

In other cases, if the Waiter was awaited, it _would_ end the test when the promise returned, but ahead of the resolving Waiter being processed, resulting in a "test count exceeds plan" error.

This patch adds a test that exercises every combination of plan(), test fn returning a promise, a pending Waiter, an awaited Waiter, and an explicit end.

Fix: https://github.com/tapjs/tapjs/issues/937


isaacs:
ci failure is an ada segfault in node 20 that's been identified and will be fixed.

PR #953, opened 2023-10-09
fix: respect prototypes when mocking

npdev453:
If trying to mock class instance (object with prototype) nothing working. 

1. overrides totaly ignored (fixed by `hasOwnProperty`)
2. result object miss his __proto__ after "cloning"

```ts

class InjectableClass {
    fn1() {
        return 'a';
    }
    fn2() {
        return 'b';
    }
}

void t.test('Encoder', (t) => {
    const mockedInjectableInstance = t.createMock(new InjectableClass(), {
        fn1() {
            return '123457';
        },
    });
})
```

mockedInjectableInstance will:
```ts
 {}
```
Instead:
```ts
 {
    fn1: () => '123457';
    __proto__: InjectableClass
}
```


isaacs:
Hm, I can see how this would make `t.createMock()` more generally useful. I've been using it just to override imports, like:

```ts
import * as FS from 'node:fs'
await t.mockImport('../src/some-file.js', {
  fs: t.createMock(FS, {
    symlinkSync: mockSymlinkSync,
  }
})
```

But the name does imply that you can use it for mocking any random object, so that makes sense.

Needs some tests added to [`src/mock/test/index.ts`](https://github.com/tapjs/tapjs/blob/main/src/mock/test/index.ts#L114) to show the changed functionality. (Something that fails without the patch, passes with the patch, and brings coverage to 100%.)

src/mock/src/index.ts
```diff
@@ -88,9 +88,15 @@ export class TapMock {
           return [k, v]
         })
         .concat(
-          Object.entries(overrides).filter(([k]) => !(k in bases))
+          Object.entries(overrides).filter(([k]) => !(Object.hasOwnProperty(bases, k))
```
isaacs:
```suggestion
          Object.entries(overrides).filter(([k]) => !Object.hasOwnProperty(bases, k))
```

src/mock/src/index.ts
```diff
@@ -88,9 +88,15 @@ export class TapMock {
...
         )
     ) as MockedObject<B, O>
+
+    if (Object.getPrototypeOf(bases) !== Object.prototype) {
+        Object.setPrototypeOf(mockedObject, Object.getPrototypeOf(bases))
```
isaacs:
```suggestion
      Object.setPrototypeOf(mockedObject, Object.getPrototypeOf(bases))
```

leonardoadame:
793c1c09f7c1262e9bf0947cf3915b6b13ab9c49

PR #955, opened 2023-10-10
fix homepage layout

Gpx:
The homepage layout looks broken right now:
node-tap

I attempted fixing it by removing the divider and using a straight gray line instead. I also adjusted the spacing for the CTA button:

Screenshot 2023-10-10 at 13 38 08


isaacs:
Whoa, weird. Works fine in Safari, but chrome is freaking out.

isaacs:
Fixed by just removing the separator. Chrome apparently is also calculating the flex widths differently, so made them explicitly 50%.

Issue #989, opened 2023-12-29
[BUG] Cannot `tap plugin add` on Windows


utybo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Have you read the `CONTRIBUTING` guide on posting bugs, and `CODE_OF_CONDUCT`?

- [X] yes I read the things

### This issue exists in the latest tap version

- [X] I am using the latest tap

### Description

You can't use `tap plugin add` on Windows. You can still add plugins manually.

Fails with `npm ENOENT`, simply because it's `npm.cmd` on Windows.

Solution: use `cross-spawn` instead of Node's built-in spawn.

### Reproduction

- Run  `tap plugin add @tapjs/clock`

### Environment

```
which tap
npm ls tap
tap versions
tap config list
tap plugin list
# if using typescript:
npm ls @isaacs/ts-node-temp-fork-for-pr-2009
npm ls typescript
# posix:
uname -a
# windows (in powershell)
"$([Environment]::OSVersion.VersionString) $(('x86', 'x64')[[Environment]::Is64BitOperatingSystem])"
```

utybo:
Actually, we don't make the call to `spawn` ourselves here as it's passed off to `foreground-child`. This is probably an upstream issue then? or maybe we should be doing a manual check ourselves there.

PR #990, opened 2023-12-29
fix(test): use cross-spawn to make npm on windows work

### LINKED ISSUE #988: [BUG] Unable to generate test class on Windows (test class generation script bugs)
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Have you read the `CONTRIBUTING` guide on posting bugs, and `CODE_OF_CONDUCT`?

- [X] yes I read the things

### This issue exists in the latest tap version

- [X] I am using the latest tap

### Description

There are two issues in the `src/test/scripts/build.mts` that lead to being unable to generate the test class on Windows:

- The call to `npm run prepare` fails silently if it fails
- On platforms that do not have a literal `npm` available on the path (i.e. Windows, which has a `npm.cmd` instead), the script fails with a ENOENT, which is never shown to the user

Because a) the fail is silent and b) there is an existing test-built in the packages distributed on npm, this leads to the user being able to run tests but unable to install 3rd-party plugins.

The fixes are respectively:

- Add a check on the result of `spawnSync` and fail if the script if something happens
- Use `cross-spawn` instead of Node's spawn, or manually add a `.cmd` if on Windows

### Reproduction

Main symptom:

- On Windows Install a plugin via your package manager, add it to your config manually
  - You can't run `tap plugin add` due to a separate issue (#989)
- Run `tap build`
- Run `console.log(t.pluginSignature)` in a test file

Your plugin will not be there.

### Environment

On Windows, latest tap

utybo:
Fixes #988 

Use cross-spawn instead of the regular Node spawn. No tests added because the tests were already failing on Windows.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1596,6 +1596,14 @@
         "node": "^16.14.0 || >=18.0.0"
       }
     },
+    "node_modules/@types/cross-spawn": {
+      "version": "6.0.6",
+      "resolved": "https://registry.npmjs.org/@types/cross-spawn/-/cross-spawn-6.0.6.tgz",
+      "integrity": "sha512-fXRhhUkG4H3TQk5dBhQ7m/JDdSNHKwR2BBia62lhwEIq9xGiQKLxd6LymNhn47SjXhsUEPmxi+PKw2OkW4LLjA==",
+      "dependencies": {
+        "@types/node": "*"
+      }
+    },
     "node_modules/@types/diff": {
       "version": "5.0.5",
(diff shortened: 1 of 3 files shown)


isaacs:
Is cross-spawn still needed? I feel like the issue in #989 is probably just a matter of missing a `shell: true` somewhere.

isaacs:
Running #958 with an update to use `shell: true` in a few places, if that passes, I'll land that and close this one. Better to save the extra dep, if possible. Keeping this open until then, since it might still be needed, should know soon.

PR #1011, opened 2024-03-19
Update README.md

bn-l:


isaacs:
Documentation for vitest belongs in vitest, not here.

PR #1039, opened 2024-09-27
Fix broken "skipping-everything" link in tap-parser

Krinkle:
Ref https://testanything.org/tap-version-14-specification.html#skipping-everything.


mbtools (approved):
LGTM 👍

PR #1046, opened 2025-01-13
Fix for Issue 1044 🐞 Unable to run tap build or install plugins when using a private registry with authentication

vala84no:
Removed lookup of registry as `pacote` already handles getting the npm registry better (it includes authentication). Removing the specific registry config passed to pacote allows it to correctly handle private repositories. I have also removed the tests associated with this lookup.

Fix for: https://github.com/tapjs/tapjs/issues/1044

### DIFF
--- a/src/run/src/plugin.ts
+++ b/src/run/src/plugin.ts
@@ -86,7 +86,11 @@ export const plugin = async (
   }
 }
 
-const add = async (args: string[], config: LoadedConfig, quiet = false) => {
+const add = async (
+  args: string[],
+  config: LoadedConfig,
+  quiet = false,
+) => {
   const error = quiet ? () => {} : console.error
   const log = quiet ? () => {} : console.log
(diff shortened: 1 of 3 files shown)


src/run/src/plugin.ts
```diff
@@ -86,7 +86,11 @@ export const plugin = async (
   }
 }
 
-const add = async (args: string[], config: LoadedConfig, quiet = false) => {
+const add = async (
+  args: string[],
+  config: LoadedConfig,
+  quiet = false,
+) => {
```
vala84no, line 93: This change is unrelated, it's simply the outcome of running `npm run format` as requested in the contribution instructions.

isaacs: Ah, yeah, that was an oversight on my part. It's already landed in main, so it should just get squashed out anyway :)

PR #1065, opened 2026-04-02
Avoid shell execution for child processes in npm and reporter logic

RinZ27:
Refactored the way child processes are spawned when installing plugins or handling custom reporters to avoid using `shell: true`. The current implementation in `npm.ts` and `handle-reporter.ts` relies on the shell to resolve executables and arguments, which can lead to unexpected behavior or potential argument injection if input contains special shell characters.

Switching to direct execution ensures that arguments are passed as a clean array to the target process. For the `npm` wrapper, a platform-specific check has been added to correctly resolve `npm.cmd` on Windows, maintaining compatibility without the overhead or risks of a shell environment. These changes improve the robustness of the CLI when dealing with various plugin names and reporter arguments. I verified that using `which(reporter)` already provides the absolute path, so bypassing the shell is safe and efficient.

### DIFF
--- a/src/run/src/handle-reporter.ts
+++ b/src/run/src/handle-reporter.ts
@@ -73,7 +73,7 @@ export const handleReporter = async (t: TAP, config: LoadedConfig) => {
     const exe = await which(reporter)
     const rargs = config.get('reporter-arg') || []
     const proc = spawn(exe, rargs, {
-      shell: true,
+      shell: false,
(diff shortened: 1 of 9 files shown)


isaacs:
Thanks! This looks great. I'm just spinning on it to update the test snapshots and get a weird new node 25 issue to stop failing, then I'll publish the fix.

PR #1066, opened 2026-04-15
Fix formatting for JSON output paths in documentation.

Potherca:
Two entries in the table in the "`.tap/report`" section of `src/docs/content/dot-tap-folder.md` where missing backticks <code>`</code>, which caused incorrect rendering:

screenshot

This MR fixes that by adding the missing backticks.


isaacs:
Landed, shipping to website now. Thanks!

PR #1070, opened 2026-09-20
parser: don't emit finish during write() (Fixes #1045)

dyk1454683243-sudo:
Fixes #1045.

## Cause

`tap-parser` is a custom `EventEmitter` writable. `end()` completed the TAP parse and then immediately `emit('finish')` / `emit('close')`, even when `end()` was invoked reentrantly from a `'plan'` listener during `write()`.

Node's `stream.Writable.toWeb()` uses `stream.finished()`. If `'finish'` fires while a web-stream write is still on the stack — before the web writable's `close()` has started — the adapter treats that as an unexpected close and rejects with `AbortError` (`ABORT_ERR`). That matches the reproduction: `readable.pipeTo(Writable.toWeb(parser))` plus `parser.end()` in the `'plan'` handler (the `@tapjs/tap-finished` pattern). Classic `parser.write()` / `parser.end()` does not go through that adapter, so it did not throw.

A second adapter bug shows up when `'finish'` is emitted synchronously from `end()` *called by* `Writable.toWeb` `close()`: the adapter clears its close-promise resolver before `close()` can return it (`undefined.promise`).

Manually emitting `'finish'`/`'close'` is also out of line with Node stream invariants (see [nodejs/help#2412](https://github.com/nodejs/help/issues/2412)).

## Fix

Keep parser completion (`'complete'`, results, comments) on `end()` as before. Stream `'finish'`/`'close'` are emitted once, with these rules:

- On the **root** parser, if `end()` runs during `write()` (e.g. `'plan'` → `p.end()`), do not emit `'finish'`/`'close'` yet. A later non-reentrant `end()` (including the adapter's `close()`) emits them.
- **Child** parsers still emit `'finish'`/`'close'` immediately, including during `write()`, so existing event logs/snapshots stay stable.
- When Node's `finished()` `onfinish` listener is attached (Writable.toWeb), emit `'finish'`/`'close'` on `nextTick` so `close()` can return its promise.
- Classic `p.end()` / `p.end(chunk)` still emit `'complete'`, then `'finish'`/`'close'` synchronously.

## Tests

`src/parser/test/to-web.ts` covers:

- Classic `write()`/`end()` still emits `complete` → `finish` → `close`.
- Classic `end(chunk)` with a `'plan'` listener that calls `end()` still emits finish/close (outer `end()` is not in a write).
- `Writable.toWeb(parser).getWriter().close()` resolves (no AbortError / TypeError).
[...]

PR #1071, opened 2026-09-20
mocha-globals: beforeEach/afterEach only wrap it() tests (Fixes #1008)

### LINKED ISSUE #1008: [BUG] mocha-globals beforeEach hooks run for describes too when it should only for it tests
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Have you read the `CONTRIBUTING` guide on posting bugs, and `CODE_OF_CONDUCT`?

- [X] yes I read the things

### This issue exists in the latest tap version

- [X] I am using the latest tap

### Description

In previous versions, nested `describe` scopes would trigger outer beforeEach/afterEach hooks only around the inner `it` tests. In 18 however it seems to _also_ trigger around the `describe` calls itself, which results in some very strange issues when used with libraries like sinon where you would typically apply a stub in the `beforeEach` and restore it in the `afterEach`. The way it works _now_ though, the hook will run an additional time on nested tests and then crash because it's trying to reapply patches that are already applied.

### Reproduction

```js
require('tap')
require('@tapjs/mocha-globals').globalize()

describe('foo', () => {
  beforeEach(() => process._rawDebug('beforeEach'))
  afterEach(() => process._rawDebug('afterEach'))

  describe('bar', () => {
    describe('baz', () => {
      it('1', () => {})
      it('2', () => {})
    })
  })
})
```

Expected output:

```
beforeEach
afterEach
beforeEach
afterEach
```

Actual Output:

```
beforeEach
beforeEach
beforeEach
afterEach
beforeEach
afterEach
afterEach
afterEach
```

### Environment

```
./node_modules/.bin/tap
dd-trace@6.0.0-pre /Users/stephen.belanger/Code/dd-trace-js
└── tap@18.7.0

tap: 18.7.0
"@tapjs/config": 2.4.15
"@tapjs/core": 1.5.0
"@tapjs/run": 1.5.0
"@tapjs/stack": 1.2.7
"@tapjs/test": 1.4.0
tap-parser: 15.3.1
tap-yaml: 2.2.1
tcompare: 6.4.5
plugins:
  "@tapjs/after": 1.1.18
  "@tapjs/after-each": 1.1.18
  "@tapjs/asserts": 1.1.18
  "@tapjs/before": 1.1.18
  "@tapjs/before-each": 1.1.18
  "@tapjs/filter": 1.2.18
  "@tapjs/fixture": 1.2.18
  "@tapjs/intercept": 1.2.18
  "@tapjs/mocha-globals": 1.1.18
  "@tapjs/mock": 1.3.0
  "@tapjs/node-serialize": 1.3.0
  "@tapjs/snapshot": 1.2.18
  "@tapjs/spawn": 1.1.18
  "@tapjs/stdin": 1.1.18
  "@tapjs/typescript": 1.4.0
  "@tapjs/worker": 1.1.18
# vim: set filetype=yaml :

# from .taprc
allow-incomplete-coverage: true
color: true
disable-coverage: true
node-arg:
  - --expose-gc
plugin:
  - "@tapjs/mocha-globals"

# env, cli, and defaults
coverage-report:
  - text
exclude:
  - "**/@(fixture*(s)|dist)/**"
include:
  - "**/@(test?(s)|__test?(s)__)/**/*.@(js|cjs|mjs|tap|cts|jsx|mts|ts|tsx)"
  - "**/*.@(test?(s)|spec).@(js|cjs|mjs|tap|cts|jsx|mts|ts|tsx)"
  - "**/test?(s).@(js|cjs|mjs|tap|cts|jsx|mts|ts|tsx)"
jobs: 8
mocha-globals: true
reporter: base
snapshot-clean-cwd: true
timeout: 30
@tapjs/after
@tapjs/after-each
@tapjs/asserts
@tapjs/before
@tapjs/before-each
@tapjs/filter
@tapjs/fixture
@tapjs/intercept
@tapjs/mock
@tapjs/node-serialize
@tapjs/snapshot
@tapjs/spawn
@tapjs/stdin
@tapjs/typescript
@tapjs/worker
@tapjs/mocha-globals
Darwin COMP-JX471G9FQQ 23.3.0 Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:44 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6000 arm64 arm Darwin
```

dyk1454683243-sudo:
Fixes #1008.

## Cause

`@tapjs/mocha-globals` maps both `describe()` and `it()` to `t.test()`. Mocha `beforeEach`/`afterEach` were wired straight through to `t.beforeEach()`/`t.afterEach()`, which run for **every** child test — including nested `describe()` suites.

That extra wrap around suites is not mocha-compatible. It also breaks libraries like sinon that stub in `beforeEach` and restore in `afterEach`, because the hook re-applies an already-applied patch when a nested `describe` starts.

Reproduction from the issue (nested `describe` + two `it` tests) produced:

```
beforeEach
beforeEach
beforeEach
afterEach
beforeEach
afterEach
afterEach
afterEach
```

Expected mocha-compatible output is one pair per `it` only:

```
beforeEach
afterEach
beforeEach
afterEach
```

Isaac noted the difficulty in the issue: `describe()` and `it()` are both `t.test()`, and tap `beforeEach` cannot distinguish them.

## Fix

Minimal mocha-globals-only change:

1. Mark `it()` tests with `tapMochaTest: true` on the child options. That key is already stripped from TAP YAML by `@tapjs/core` (`/^tapMochaTest/` in `clean-yaml-object`).
2. Wrap mocha `beforeEach`/`afterEach` so the user hook runs only when the child is an `it()` test. Nested `describe()` suites still inherit the hook list (so inner `it()` tests still get parent hooks), but the hook is a no-op for the suite itself.

`before()`/`after()` (once-per-suite) are unchanged.

## Tests

- New `src/mocha-globals/test/nested-hooks.ts` is the issue reproduction: nested `describe` scopes and an exact log assertion.
- Existing `hooks.ts` snapshot updated: root `beforeEach`/`afterEach` no longer fire for the suite itself or for the log-check `describe`.

`@tapjs/mocha-globals` suite: 63/63 pass.

No existing open/recent PR claimed this fix (searched tapjs/tapjs for #1008, mocha-globals, beforeEach). This does not collide with open #1070 (parser toWeb AbortError).


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

